PR #61, opened 2020-06-11
Fix deployDev workflow on Windows

mdial89f:
The deployDev workflow builds a local environment with mounted folders and live reloading of the react container.
This was broken on Windows.
Why it was broken just on Windows isn't known, and probably won't be.
In investigating this issue, a better pattern for building and running the dev react container was found.  

We're moving the npm install from the image run time to the image build time.  By using a data volume in addition to the react bind mount, it seems we can mount the react directory without the host node_modules overwriting the containers.
See here:  https://stackoverflow.com/questions/30043872/docker-compose-node-modules-not-present-in-a-volume-after-npm-install-succeeds
The key piece here is in the top answer.  "A workaround is to use a data volume to store all the node_modules, as data volumes copy in the data from the built docker image before the worker directory is mounted. "
Our data volume is /react/node_modules.

So, I really am not sure why npm install gets hungup when executing on a node container on windows.  But it doesn't matter... with this change, we swerve that issue entirely.


mdialcollabralinkcom:
I opened this (again) against the wrong branch... we should update the default branch to master

PR #106, opened 2020-06-24
first coming moving from a single api folder to a folder for each api…

dwhitestratiform:
…, also adding support for docker image for each api


dwhitecl:
this work moved branches and this PR is no longer valid

PR #121, opened 2020-06-26
Add writeup for the temporary URLs feature, as per issue #94.

tadhg-ohiggins:
Edit by mail: bad—
Temporary URLs
Would be much cleaner

Add a docs directory and the writeup for temporary URLs from #94.


cannandev (approved):
Nice work!

PR #182, opened 2020-07-10
Feature/date range

### LINKED ISSUE #158: Create Date Range component to capture dates
Let's use four fields (two `<DateField>` components) to capture dates throughout the application. We already have this in section 2B. Let's make this into a reusable component.

For example, Section 2A - question 3b. Instead of a textfield, use the date range component. As a child element, It should be styled as a conditional  (blue border-left)

Screen Shot 2020-07-02 at 2 03 32 PM

alewoo28:
-Created two component 'DateOfRangeComponent' and 'DateRange'
-DateOfRangeComponent validates all date input, and uses text fields to build a date component 
-DateRange validates that the dates are chronological
-Both components can be customized later through props for nice, repeatable code
-Replaced CMS design system DateField component in 2b with custom component so that the date may be omitted

### DIFF
--- a/app/react/src/components/layout/DateOfRangeComponent.js
+++ b/app/react/src/components/layout/DateOfRangeComponent.js
@@ -19,6 +19,7 @@ class DateOfRangeComponent extends Component {
 
       yearEndErr: false,
       yearEnd: "",
+      dummyDigit: 10,
     };
     this.validateMonth = this.validateMonth.bind(this);
     this.validateYear = this.validateYear.bind(this);
@@ -120,7 +121,11 @@ class DateOfRangeComponent extends Component {
                     ),
                     this.props.validateDateRange)
                   }
-                  value={this.state.monthStart}
+                  value={
+                    this.props.previousEntry === true
+                      ? this.state.dummyDigit - 5
+                      : this.state.monthStart
+                  }
                 />
                 <div className="ds-c-datefield__separator">/</div>
                 <TextField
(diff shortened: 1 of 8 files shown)


jgillis1 (approved):

app/react/src/components/layout/DateOfRangeComponent.js
```diff
@@ -0,0 +1,212 @@
...
+            <div className="date-range-start">
+              <h3 className="question-inner-header"> Start </h3>
+              <div className="ds-c-field__hint"> From mm/yyyy to mm/yyyy</div>
+              <div className="date-range-start-wrapper">
+                <TextField
+                  className="ds-c-field--small"
+                  errorMessage={this.state.monthStartErr}
+                  name="monthStart"
+                  numeric
+                  onChange={this.validateMonth}
```
cannandev, line 115: Let's change the validation to happen onBlur instead of onChange. That way, the user doesn't have to see the error message while they're still typing. e.g. `07`

app/react/src/components/layout/DateOfRangeComponent.js
```diff
@@ -0,0 +1,212 @@
...
+
+  render() {
+    return (
+      <Fragment>
+        {this.props.range ? (
+          <div className="date-range">
+            <div className="date-range-start">
+              <h3 className="question-inner-header"> Start </h3>
+              <div className="ds-c-field__hint"> From mm/yyyy to mm/yyyy</div>
+              <div className="date-range-start-wrapper">
```
cannandev, line 109: The errors should stack at the beginning of the `date-range-start-wrapper`, or as close to it as possible, spanning the width of the month and year fields. See example attached
Screen Shot 2020-07-13 at 4 37 51 PM

cannandev (commented):
See my comments in `app/react/src/components/layout/DateOfRangeComponent.js`

PR #235, opened 2020-07-24
Refactor project

mdial89f:
This is a fairly large set of changes derived from work done elsewhere.
I'll do my best to make a list here.  
- postgres and our microservices have been split into two distinct top level layers... data and frontend
- these two layers are deliberately (and in our case pretty aggressively) decoupled
- the data layer now includes a postgres_deployer service that deploys our data code to postgres
- ECR repository url's are looked up in terraform, rather than fed through CasC and jenkins variables
- Passwords are no longer maintained anywhere in Jenkins... AWS SSM is used throughout
- Terraform code now lives with its layer.  The top level 'infra' folder is gone, and there is instead an 'aws' terraform dir inside each layer (data, frontend)
- The master branch build is now represented by a build called 'master'.  It's viewable directly on the Jenkins homepage.
- The master pipeline and all higher environments now have some logic to skip stages as needed.  In other words, if no data code has been updated, the data layer won't redeploy.  Likewise for the frontend layer.
- The 'dev' environments are still built by the multibranch 'dev' pipeline, viewable from the jenkins homepage.
- The 'dev' pipeline searches for branches that match the pattern 'dev-*'. This is an update from the previous 'master-*'.
- Dev pipelines now have a list of options when you click Build Now with Parameters.  You may destroy/deploy/test in any combination.
- local deployments can still be affected by running deploy.sh at the top level.  
- The difference now with local deployments is that each layer is deployed in series.
- The top level deploy.sh simply calls data/deploy.sh and frontend/deploy.sh
- You may run individual layer's deploy.sh instead of the top level deploy.sh.
- This means that a React developer can deploy the whole stack, but then go into the frontend directory and deploy/destroy/deploy/etc that layer at will, leaving the data layer unaffected.
- Theres a top level test.sh.  Run sh test.sh to run our test suite.  it's as basic as it comes, but verifies the networking.  Jenkins runs this same suite.

I know these seem pretty aggressive.  
I felt it was important to very explicitly decouple our layers.
Bolting new independent layers to these environments is now straightforward.
Those new layers need not be in this repositroy; that's key.  We can support another team developing and deploying a new microservice or layer on their own timeline in their own reopsitory.
We can also break off our data layer from our frontend layer if we want.
So if it seems like this data layer and frontend could live in separate repositories... you've nailed it.  By design.

Your code is moving.   app/react is now frontend/react.
The data code moved from app/postgres to data/postgres_deployer, but I'm not aware of anyone developing data code yet.

Existing work in flight can be updated with these changes once they're on master.  Git will see it as a moved file, not a delete/add.

Please send me questions.  I know there will be some, along with some issues.  

Thanks


mdial89f:
Please squash when merging

dwhitestratiform (approved):

PR #321, opened 2020-08-07
Trace question number starting from end of string

andrewadcock:
Split the question number into an array and begin parsing from last item


tadhg-ohiggins:
There are some sections in here that get array length and then use it to access a point in the array that can be replaced by using `slice`.

```
let idArray = this.props.id.split("-");
let length = idArray.length;
let questionNumber = Number(idArray[length - 1]) + ":";
```

Could be something like:

`let questionNumber = Number(this.props.id.split("-").slice(-1)) + ":"`

andrewadcock:
Tightened it up a bit, based on your suggestion

jgillis1 (approved):

PR #325, opened 2020-08-07
Add fallback support for section1 prototype

andrewadcock:
Adds support for the legacy /section1

### DIFF
--- a/docs/section-schemas/backend-json-section-3.json
+++ b/docs/section-schemas/backend-json-section-3.json
@@ -43,8 +43,13 @@
                         "hide_if": {
                           "target": "$..*[?(@.id=='2020-03-a-01-01')].answer.entry",
                           "values": {
-                            "interactive": [null, "no"],
-                            "noninteractive": ["no"]
+                            "interactive": [
+                              null,
+                              "no"
+                            ],
+                            "noninteractive": [
+                              "no"
+                            ]
                           }
                         }
                       }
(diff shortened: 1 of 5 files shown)


andrewadcock:
Updated with changes suggested by @tadhg-ohiggins

docs/section-schemas/backend-json-section-3.json
```diff
@@ -409,18 +382,15 @@
                 ],
                 "context_data": {
                   "conditional_display": {
+                    "toMatch": { "no": "2020-03-d-01-01" },
```
tadhg-ohiggins, line 385: Can you remove this line? The JSON won't validate if it's there. @andrewadcock

tadhg-ohiggins (changes requested):
Overall, I think that most of these changes look like whitespace tweaks, and I think we should probably revert those. I think that ideally this PR shouldn't need any changes to the JSON docs; let me know if there is a necessary change that I've missed.

andrewadcock:
The whitespace changes came from my linter. I reverted back to the original though.

tadhg-ohiggins:
I'm still seeing a `toMatch` in there; should be good to go once they're all removed.

andrewadcock:
I believe I did it right this time. All the toMatch should be removed.

tadhg-ohiggins (approved):

PR #334, opened 2020-08-10
Okta Authentication

sethsacher:
Okta-based authentication wrapping the entire CARTS React frontend. If a user is not authenticated, they will be directed to the default CMS Okta login screen. After providing valid credentials, the user is directed to the CARTS homepage. This is currently only implemented for local development, and only for the frontend. Additional checks should be implemented on the Django backend to ensure an unauthenticated user cannot access an API endpoint. 

A test /profile route has been added to show the user attributes accessible, including the CMS Roles. This is not an exhaustive list of user profile attributes.


sethsacher:
Replaced by https://github.com/18F/cms-carts-seds/pull/337

PR #335, opened 2020-08-11
JSON for Section 3C

tadhg-ohiggins:
Demote percentage—
It is now a mere action.
These docs are fluid.

Remove `percentage` as a `fieldset_type` option, instead make it an `action` for `synthesized_value` (and by extension `synthesized_table`). Add example showing its use.

This PR now encompasses a lot more, so:

### Section 3C

In Section 3C
Are the worst questions to code—
At least I hope so.

Add JSON for Section 3C, including along the way:

- Add `comment` property to `fieldset`.
- Add `unmarked_descendants` value for `fieldset_type`.
- Add `marked` value for `fieldset_type`.
- Add `checkbox_flag` question type.
- Add boolean type support in `conditional_display`.
- Add documentation for all of the above.

### DIFF
--- a/docs/section-schemas/backend-json-section-3.json
+++ b/docs/section-schemas/backend-json-section-3.json
@@ -781,6 +781,7 @@
                       },
                       {
                         "type": "fieldset",
+                        "fieldset_type": "datagrid",
                         "context_data": {
                           "skip_text": "1b–1e omitted because you indicated that you don't have the age range breakdown data available.",
                           "conditional_display": {
@@ -887,6 +888,7 @@
                       },
                       {
                         "type": "fieldset",
+                        "fieldset_type": "datagrid",
                         "context_data": {
                           "skip_text": "2b–2e omitted because you indicated that don't you have the age range breakdown data available.",
                           "conditional_display": {
@@ -985,6 +987,7 @@
                       },
                       {
                         "type": "fieldset",
+                        "fieldset_type": "datagrid",
                         "context_data": {
                           "skip_text": "3b–3e omitted because you indicated that you don't have the age range breakdown data available.",
                           "conditional_display": {
@@ -1079,6 +1082,7 @@
                               },
                               {
                                 "type": "fieldset",
+                                "fieldset_type": "datagrid",
                                 "context_data": {
                                   "skip_text": "3f 2–5 omitted because you indicated that you don't have the age range breakdown data available.",
                                   "conditional_display": {
@@ -1182,6 +1186,7 @@
                       },
                       {
                         "type": "fieldset",
+                        "fieldset_type": "datagrid",
                         "context_data": {
                           "skip_text": "4b–4e omitted because you indicated that you don't have the age range breakdown data available.",
                           "conditional_display": {
@@ -1276,6 +1281,7 @@
                               },
                               {
                                 "type": "fieldset",
+                                "fieldset_type": "datagrid",
                                 "context_data": {
                                   "skip_text": "4f 2–5 omitted because you indicated that you don't have the age range breakdown data available.",
                                   "conditional_display": {
@@ -1397,6 +1403,7 @@
                       },
                       {
                         "type": "fieldset",
+                        "fieldset_type": "datagrid",
                         "context_data": {
                           "skip_text": "6b–6e omitted because you indicated that don't you have the age range breakdown data available.",
                           "conditional_display": {
@@ -1495,6 +1502,7 @@
                       },
                       {
                         "type": "fieldset",
+                        "fieldset_type": "datagrid",
                         "context_data": {
                           "skip_text": "7b–7e omitted because you indicated that you don't have the age range breakdown data available.",
                           "conditional_display": {
@@ -1589,6 +1597,7 @@
                               },
                               {
                                 "type": "fieldset",
+                                "fieldset_type": "datagrid",
                                 "context_data": {
                                   "skip_text": "7f 2–5 omitted because you indicated that you don't have the age range breakdown data available.",
                                   "conditional_display": {
@@ -1670,7 +1679,7 @@
                           "entry": null
                         },
                         "context_data": {
-                          "skip_text": "4a omitted because you indicated that you have the age range breakdown data available.",
+                          "skip_text": "8a omitted because you indicated that you have the age range breakdown data available.",
                           "conditional_display": {
                             "type": "conditional_display",
                             "comment": "Interactive: Hide if 2020-03-c-04-01-a is false or null; noninteractive: same.",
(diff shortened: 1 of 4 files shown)


docs/section-schemas/backend-json-section-3.json
```diff
@@ -301,11 +301,2067 @@
...
+                    "questions": [
+                      {
+                        "comment": "Yes, this is a marked fieldset, i.e. a fieldset mainly pretending to be a question for the purposes of the list markers, immediately followed by an unmarked_descendants fieldset used so that the age range breakdown question gets to pretend to not be a question for the purposes of the list markers.",
+                        "type": "fieldset",
+                        "fieldset_type": "unmarked_descendants",
+                        "questions": [
+                          {
+                            "id": "2020-03-c-04-01-unmarked_descendants",
+                            "label": "I don’t have data for the individual age groups. I’ll report data for the total number for all age groups (0-18 years) instead.",
+                            "type": "checkbox_flag",
```
cannandev, line 742: I left a comment in the google doc about the location of the checkbox. From an accessibility standpoint, it should belong to part 4, not question 1. I'd like to move it outside of the question fieldset. I propose two ways to handle this:
id: `2020-03-c-04-a`, which wouldn't conflict with numbering for the inputs that follow. 
**or** 
id: `2020-03-c-04-01` and continue with that sequence for the rest of the structures that follow, but we have an additional `marker` property that we'll use to number each question instead of the `id`

docs/section-schemas/backend-json-section-3.json
```diff
@@ -301,11 +301,2067 @@
...
+                        "comment": "This is the first real question so far in this part…",
+                        "type": "question",
+                        "id": "2020-03-c-04-01-a",
+                        "label": "Total for all ages",
+                        "type": "integer",
+                        "answer": {
+                          "entry": null
+                        },
+                        "context_data": {
+                          "skip_text": "1a omitted because you indicated that you have the age range breakdown data available.",
```
cannandev, line 762: Do we need to show this text? The user won't notice the missing fields once they check the checkbox.

docs/section-schemas/backend-json-section-3.json
```diff
@@ -301,11 +301,2067 @@
...
+            "questions": [
+              {
+                "type": "fieldset",
+                "label": "January – March 2019 (start of the cohort)",
+                "questions": [
+                  {
+                    "comment": "The “sum” question, that they answer if they don't have the breakdowns, can't really be a true top-level question, because then we couldn't hide it in the case where the breakdown data is available. So it has to be a subquestion, meaning that the top-level question isn't real and can't actually be answered. This introduces the need for the special fieldset used here; see the schema docs for more details.",
+                    "type": "fieldset",
+                    "fieldset_type": "marked",
```
cannandev, line 727: Instead of marked/unmarked, I propose an `input_grid` structure with the following properties:
```
            id="2020-03-c-04-04"
            title="July - September of 2020 (18 months later)"
            label="How many children were no longer enrolled in CHIP six months later?"
            hint=["Possible reasons for no longer being enrolled","Transferred to another health insurance program other than CHIP","Didn’t meet eligibility criteria anymore","Didn’t complete documentation","Didn’t pay a premium or enrollment fee"]
            isChild="false" />
```
```
            id=2020-03-c-04-04-a`
            title=""
            label="How many children were no longer enrolled in CHIP but were enrolled in Medicaid six months later?"
            hint=""
            isChild="true" 
```
I'd also love a `marker` property that just said `4` or `4a`, to avoid substring-ing the id, but that might be asking too much.

`isChild` would let us style this question differently - in this case, indented from 4. Note that from a structural standpoint, there is no parent-child relationship allowed with this structure.

cannandev (changes requested):

docs/section-schemas/backend-json-section-3.json
```diff
@@ -301,11 +301,2083 @@
...
+                    }
+                  }
+                ]
+              },
+
+
+
+              {
+                "type": "fieldset",
+                "label": "January – March 2020 (12 months later)",
```
cannandev, line 1363: We need a flag that this section should be inactive and filled out next year.

docs/section-schemas/backend-json-section-3.json
```diff
@@ -301,11 +301,2083 @@
...
+                      }
+                    ]
+                  }
+                ]
+              },
+
+
+              {
+                "type": "fieldset",
+                "label": "July – September 2020 (18 months later)",
```
cannandev, line 1869: We need a flag that this section should be inactive and filled out next year.

cannandev (changes requested):

docs/section-schemas/backend-json-section-3.json
```diff
@@ -301,11 +301,2067 @@
...
+                                "type": "boolean"
+                              },
+                              "entry": null
+                            }
+                          }
+                        ]
+                      },
+                      {
+                        "comment": "This is the first real question so far in this part…",
+                        "type": "question",
```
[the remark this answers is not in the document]
tadhg-ohiggins: The `type: question` part was a stray, but this should be `integer`; the answer can't be anything but a whole number.

docs/section-schemas/backend-json-section-3.json
tadhg-ohiggins: We can move it, we just have to decide to where, based on the comments in the doc. I'm not in favor of adding a `marker` property. To move this checkbox question I'd keep it in an `unmarked_descendants` `fieldset` because that will flag that it shouldn't have a marker and shouldn't alter the markers of questions after it.

docs/section-schemas/backend-json-section-3.json
tadhg-ohiggins: Depends on whether we're showing the labels, and on whether or not e.g. 3 has a subquestion. If it does, its label will be 3f, and so in order to have things make sense to the user we need to let them know why we're skipping to `f`.

docs/section-schemas/backend-json-section-3.json
tadhg-ohiggins: I don't think I understand how this would work. I do think that `input_grid` wouldn't solve the problems that `marked`/`unmarked_descendants` tackle.

docs/section-schemas/backend-json-section-3.json
tadhg-ohiggins: I left that out because I wasn't sure yet how to handle it; I don't think that consideration should hold up the PR.

docs/section-schemas/backend-json-section-3.json
cannandev: Marking this issue `Resolved` since we decided 3a will be 4, etc. No subquestion.

cannandev (approved):

PR #360, opened 2020-08-14
[Spike] Connect section API to Redux store

hbillings:
A spike containing an example pattern for:
- hitting the JSON API from Redux
- populating the Redux store with the result
- calling the Redux store from the individual section component

Most patterns taken from https://redux.js.org/advanced/async-actions (hooking up the Redux store with redux-thunk and async actions) and https://redux.js.org/advanced/example-reddit-api (dispatching actions from components).

Please take whatever is useful here and leave whatever isn't! Happy to chat more about any of this, with the caveat that I have no idea what I'm doing


frontend/api_postgres/carts/management/commands/generate_fixtures.py
```diff
@@ -16,7 +16,7 @@ def handle(self, *args, **options):
             paths = paths + [_ for _ in fd.glob(glob)]
 
         schemapath = Path(fd, "backend-section.schema.json")
-        paths = paths + schemapath
+        paths = paths + [schemapath]
```
hbillings, line 19: oops

frontend/react/src/components/sections/section1/Section1.js
```diff
@@ -12,6 +12,10 @@ import {
   TabPanel,
   TextField,
 } from "@cmsgov/design-system-core";
+import {
+  selectSection,
+  fetchSectionData
+} from "../../../store/sectionData";
```
hbillings, line 18: At some point, it might make sense to have the actions and reducers split into `actions.js` and `reducers.js` for clarity about what you're importing in files like this.

frontend/react/src/components/sections/section1/Section1.js
```diff
@@ -61,10 +77,25 @@ class Section1 extends Component {
...
-  name: state.stateUser.name,
-  programType: state.stateUser.programType,
-  year: state.global.formYear,
-});
+// Section1.propTypes = {
+//   selectedSection: PropTypes.string.isRequired,
+//   data: PropTypes.object.isRequired,
+//   isFetching: PropTypes.bool.isRequired,
+//   dispatch: PropTypes.func.isRequired
+// }
```
hbillings, line 85: I'd also recommend adding the proptypes package! It's super useful to be able to see at a glance what kinds of props your app is expecting (plus it helps with testing whether required props are coming through as expected).

frontend/react/src/components/sections/section1/Section1.js
```diff
@@ -22,6 +26,18 @@ class Section1 extends Component {
...
+    const { dispatch, selectedSection } = this.props;
+    dispatch(selectSection('1'));
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props.selectedSection !== prevProps.selectedSection) {
+      const { dispatch, selectedSection } = this.props;
+      dispatch(fetchSectionData(selectedSection));
+    }
+  }
```
hbillings, line 39: These should really happen in a `mapDispatchToProps` call instead, but I ran out of time. Will update here in a minute if I can!

frontend/react/src/store/sectionData.js
```diff
@@ -0,0 +1,91 @@
+// Mostly taken from https://redux.js.org/advanced/async-actions
+
+const initialState = {
+  selectedSection: '1',
+  dataBySection: {
+    section1: {
+      isFetching: false,
+      items: []
```
hbillings, line 8: I wasn't sure what the correct key was, so please feel free to change `items` to whatever makes sense.

frontend/react/src/store/sectionData.js
```diff
@@ -0,0 +1,91 @@
...
+      })
+    default:
+      return state
+  }
+}
+
+export const fetchSectionData = (section) => {
+  return function (dispatch) {
+    dispatch(requestSectionData(section));
+    return fetch(`http://localhost:8000/structure/2020/${section}?format=json`)
```
hbillings, line 87: This should probably use Axios or another library that polyfills fetch.

frontend/react/src/store/sectionData.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export const selectSection = (section) => {
+  return {
+    type: SELECT_SECTION,
+    section
+  }
+}
+
+export const selectedSection = (state = '0', action) => {
```
hbillings, line 40: `state` should maybe just be a blank string here

frontend/react/src/components/sections/section1/Section1.js
```diff
@@ -22,6 +26,18 @@ class Section1 extends Component {
     };
   }
 
+  componentDidMount() {
+    const { dispatch, selectedSection } = this.props;
+    dispatch(selectSection('1'));
```
hbillings, line 31: I'd highly recommend considering a functional approach with hooks across the board, but I went with class-based components because it seemed like the rest of teh codebase followed that pattern.

tadhg-ohiggins:
@jgillis1 @alewoo28 Could you have a look at this to see how Heather's approach would fit in with the steps you've already taken to pull data into React from the Django API?

frontend/react/src/store/sectionData.js
alewoo28: Yeah, I agree Axios is a good idea

frontend/react/src/components/sections/section1/Section1.js
alewoo28: Hmm, why are we importing selectSection & fetchSectonData this way instead of through mapStateToProps?

Agreed that we should split the reducers and actions up for clarity, that makes sense to me

frontend/react/src/components/sections/section1/Section1.js
alewoo28: Okay cool! Thanks Heather

frontend/react/src/components/sections/section1/Section1.js
alewoo28: Good idea! That package should already be in the modules, I think i use it in DateRange.js, it just hasnt been implemented across the board yet.

frontend/react/src/store/sectionData.js
alewoo28: This is great. Do you think we should have a separate file/reducer for each section? I thought maybe it could make things clearer, lmk what you think about that

alewoo28 (approved):
Looks awesome. This is pretty much exactly how I thought it would be, thank you Heather!

jgillis1 (approved):

PR #361, opened 2020-08-17
Feature/temporary url

cannandev:
@jgillis1 @andrewadcock @alewoo28 I'm using React Hooks (`useState`, `useRef`) and function components for the share link modals. I also used named exports to avoid any mistakes on import. 

Please review and let me know any questions you have compared with the `Class` approach. I can join one of your coworking sessions this week to discuss.


jgillis1:
I thought we had agreed last week that we were not going to use hooks.

cannandev:
Replacing with new PR #403  with fewer commits.

PR #376, opened 2020-08-19
Set up OKTA Enabled and OKTA Disabled options

jgillis1:
Set up OKTA Enabled and OKTA Disabled options in App.js and Header.js so we can turn on/off authentication for development purposes until EUA/LDAP Groups are created


cannandev (changes requested):
The essential code `<Router><Route component={Home} /></Router>` is really hard to read within the comments. Please a multiline comment at the beginning of the file for explanations of what to expect in that file. 
```
/**
 * 
 * @param {*} component 
 */
 ```
React follows [JSDoc standard](https://jsdoc.app/tags-param.html)

PR #389, opened 2020-08-20
Re-add Okta Authentication (with it disabled for now)

jgillis1:
Reverts 18F/cms-carts-seds#341

### DIFF
--- a/frontend/react/src/App.js
+++ b/frontend/react/src/App.js
@@ -1,22 +1,19 @@
 import React from "react";
 import "font-awesome/css/font-awesome.min.css";
 import "./App.scss";
-import Routes from "./reactRouter";
-import Header from "./components/layout/Header";
-import Footer from "./components/layout/Footer";
+import { BrowserRouter as Router, Route } from 'react-router-dom';
+import { Security, SecureRoute, LoginCallback } from '@okta/okta-react';
+import Home from "./Home";
+import config from './auth-config';
 
 function App() {
-  let VisibleHeader =
-    window.location.pathname.split("/")[1] === "reports" ? null : <Header />;
-
-  let VisibleFooter =
-    window.location.pathname.split("/")[1] === "reports" ? null : <Footer />;
   return (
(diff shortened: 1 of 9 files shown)


frontend/react/src/App.js
```diff
@@ -1,24 +1,39 @@
...
-    <div className="App" data-test="component-app">
-      <InitialDataLoad />
-      {VisibleHeader}
-      <Routes />
-      {VisibleFooter}
-    </div>
+    <InitialDataLoad />
+    <Router>
+      {/* OKTA Enabled */}
+      {/* <Security {...config.oidc}>
```
cannandev, line 29: Default: Show `<Security/>`, hide `<Route component={Home}/>`
Flag present: Hide `<Security/>`, show `<Route component={Home}/>`

frontend/react/src/components/layout/Header.js
```diff
@@ -54,6 +55,10 @@ class Header extends Component {
                       <a href="javascript:void(0)">Manage account</a>
                     </li>
                     <li className="logout">
+                      {/* OKTA Enabled */}
```
cannandev, line 58: Default: Show `<Logout/>`
Flag present: Hide `<Logout/>`

frontend/react/src/App.js
```diff
@@ -1,24 +1,39 @@
+/* 
+To Enable OKTA Authentication, you will need to:
```
cannandev, line 2: Instead of enabling/disabling by commenting out code, let's add a url parameter like `?okta_disabled=true`
Okta auth should be on by default and skipped in the presence of the flag.

cannandev (changes requested):

cannandev (commented):
@jgillis1 Did the Okta folks confirm that client IDs were public? I can't remember.

mgwalker:
Is this PR still valid? #526 and #535 also enabled Okta with a bypass mechanism, but I don't know if it covered everything that's in this PR.

PR #391, opened 2020-08-20
Load sections from the API at page load

mgwalker:
Inspired by where @hbillings started on #360.

1. Creates a component that exists solely for firing off the request to load data initially
2. Adds a redux action to load data
3. Hooks up the component's initial load so that it dispatches the load action
4. Calls the API's `GET /sections/` method and dispatches the returned data
5. Uses the `formData` reducer to push initial data into the store
6. Switches to `BascInfoApi/BasicInfo` component to dynamically render content from the API

I understand from @tadhg-ohiggins that the `/sections/` endpoint is not the right place to pull data from and the `formData` part of the state may not be the best state name, but it seems reasonable that we can figure out those details once the basic workflow is in place.

But in a nutshell, with this PR, the Basic Info section is dynamically built and populated using the data from the API.

related to #314

### DIFF
--- a/frontend/react/src/actions/initial.js
+++ b/frontend/react/src/actions/initial.js
@@ -4,7 +4,7 @@ export const LOAD_SECTIONS = "LOAD SECTIONS";
 
 export const loadSections = () => {
   return async dispatch => {
-    const { data } = await axios.get("//localhost:8000/sections/");
+    const { data } = await axios.get("//localhost:8000/v1/sections/2020/AK");
     dispatch({ type: LOAD_SECTIONS, data });
   };
 };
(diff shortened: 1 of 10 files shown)


frontend/react/src/actions/initial.js
```diff
@@ -0,0 +1,10 @@
+import axios from "axios";
+
+export const LOAD_SECTIONS = "LOAD SECTIONS";
+
+export const loadSections = () => {
+  return async dispatch => {
+    const { data } = await axios.get("//localhost:8000/sections/");
```
tadhg-ohiggins, line 7: This should include year and state. Assuming https://github.com/18F/cms-carts-seds/pull/396 gets merged, this should be:
`localhost:8000/api/v1/sections/2020/<state>`.

Or, if we're still waiting on wiring up the relevant state at that point, it should get the default sections via:
`localhost:8000/api/v1/generic-sections/2020`

tadhg-ohiggins (approved):

tadhg-ohiggins (approved):

PR #401, opened 2020-08-23
Experimental basic info

tadhg-ohiggins:
Load from API
Pass id and stuff along
dot dot dot profit!

This is my take on making Section 0 work via components passing their sub-components down a chain.
The basic approach is that every component still has `mapStateToProps` that passes the `Data` object to to it, but each parent also passes the id of the child to the child component, and then the child component uses that to look up its own data in the Data object. That's what made sense to me, but if there are better ways please bring them up.

This doesn't handle changes to the input values at all, because I'm not sure what the best way to do that is. I did create a function, `setAnswerEntry`, that I intend to be the default function for handling changes, but right now all it does is log things to console.


frontend/react/src/components/sections/basicinfoapi/Subsection.js
```diff
@@ -0,0 +1,33 @@
...
+
+  ) : null;
+}
+
+const mapStateToProps = (state, ownProps) => ({
+  fragment: selectFragmentById(state, ownProps.subsectionId),
+  subsectionId: ownProps.subsectionId,
+  abbr: state.stateUser.currentUser.state.id,
+  year: state.global.formYear,
+  Data: selectSectionByOrdinal(state, extractSectionOrdinalFromId(ownProps.subsectionId)),
```
tadhg-ohiggins, line 27: Instead of hardcoding 0 here, extract it from the id passed to the component.

frontend/react/src/components/sections/basicinfoapi/QuestionLike.js
```diff
@@ -0,0 +1,138 @@
...
+  ["text_small", QuestionTextSmall],
+  ["text_medium", QuestionTextMedium],
+  ["text_multiline", QuestionTextMultiline],
+])
+
+// Connect question types to functions via their types:
+const QuestionHolder = (Data, fragment, elementId, setAnswer) => {
+  const func = QuestionMap.has(fragment.type) ? QuestionMap.get(fragment.type) : QuestionMap.get("text");
+  return func(Data, fragment, setAnswer);
+}
```
mgwalker, line 69:
```suggestion
const QuestionHolder = ({ Data, fragment, elementId, setAnswer }) => {
  const Component = QuestionMap.has(fragment.type) ? QuestionMap.get(fragment.type) : QuestionMap.get("text");
  return (<Component Data={Data} fragment={fragment} changeFunc={setAnswer} />);
}
```

mgwalker (commented):
Some React-y-ness feedback, but no blockers for experimental dev. I like this approach. I'd suggest breaking it into a few different files for maintainability, but the basic approach meshes with what I had in my head.

I'm interested in digging in more because I'm curious how form validation plays into all this. I heard that come up in this morning's chat.

mgwalker:
Works as advertised. I think we could play around some more with how props get passed down from the `QuestionHolder` component to make the lower-level input components a little more agnostic, which should make it easier to build them out in isolation. In any case, isolating the logic for picking a component based on the question type is great.

mgwalker (approved):

Issue #402, opened 2020-08-24
Get new UUID for temporary url


cannandev:
- [ ] django API returns UUID for Data Centralizer user

PR #403, opened 2020-08-24
Feature/temporary share link

cannandev:
React component for generating, reviewing and revoking a new shared link. 

### Share View (seen by Data SME) 

- Create route for share view (seen by Data SME) at `/shared/:uuid`
- Hitting Submit dispatches `UNDER_REVIEW` action.
- Add informational text at top
- Add submit link at top and bottom
- Hide other routes, navigation/global links
- Add dialogue box that confirms submitting/thank you
- Send user to No Access page 

### Share Link and Modal (seen by Data Centralizer)

- Remove old Share Progress toggle 
- Click Generate button dispatches `CREATE_SHARE_LINK` to store. 
- Click Revoke button dispatches `EXPIRE_SHARE_LINK` to store. 

Related to #402


frontend/react/src/reactRouter.js
```diff
@@ -13,14 +13,15 @@ import Section3dapi from "./components/sections/section3dapi/Section3D";
...
+import ShareView from "./components/layout/ShareView";
 
 import Section3AApi from "./components/sections/section3Aapi/Section3A";
 
 let VisibleSidebar =
   window.location.pathname === "/" ||
-  window.location.pathname.split("/")[1] === "reports" ? null : (
-    <Sidebar />
-  );
+    window.location.pathname.split("/")[1] === ("reports" && "shared") ? null : (
```
mgwalker, line 22: What does this conditional do? Reading it, I would expect it to end up being the same as:

```js
window.location.pathname.split("/")[1] === "shared"
```

because the `("reports" && "shared")` would be evaluated to just `"shared"`. Is that the intent?

mgwalker (changes requested):
This seems fine, but with one question/fix about the conditional for whether to show the side nav.

frontend/react/src/reactRouter.js
cannandev: That was the intent. Rewrote to explicitly say that, instead of trying to be fancy.

mgwalker (approved):

mgwalker:
When I joined the project I didn't realize approvers were supposed to merge, so I didn't merge this PR... 😬  Should we still merge it now that we're punting on share links for the initial release? Or would it be better to close it and put a link to it in the issue so we don't lose track of the work for the future? (Just trying to clean up the PR list and I accidentally left this one lingering...)

mgwalker:
Going to close this PR, but will link to it in the associated issue so the work isn't lost for the future.

PR #407, opened 2020-08-24
Feature/api section3

TGriesemer:


jgillis1:
This PR is for the Questions component which holds all types of questions and answers.

alewoo28:
Section 3F:
-Receives section JSON data from redux (this.props.Data)
-Narrows it down to subsection 3F (subsectionData)
-Maps through the array of parts for each subsection
-Passes down a method to the QuestionComponent binding all questions to the parent context (sectionContext)

QuestionComponent:
-Maps through an array of questions, checking for question type
-Each question is sent to the parent section, some of them are sent after validation (specific to the question type)

alewoo28:
3F is sucessfully reading from the database via the redux store and rendering all questions via the question component

jgillis1:
@alewoo28 Could you look at merge conflicts here?

frontend/react/src/components/sections/section3Fapi/questions/Questions3FApi.js
```diff
@@ -0,0 +1,162 @@
...
+                            answer={question.answer.entry}
+                            conditional={question.conditional}
+                            children={question.questions}
+                            valueFromParent={this.state[question.id]}
+                            onChange={this.handleChange}
+                          />
+                        );
+                      }
+                    )
+                  : null}
```
jgillis1, line 73: Is there a reason we're mapping through components this way rather than going through the Questions component?

jgillis1:
Section 1 will need to be redone using the new connection method, however, I don't want to hold up this PR which has important pieces for needed in it to proceed forward in other areas.

frontend/react/src/components/sections/section3Fapi/questions/Questions3FApi.js
andrewadcock: This was a tough decision. We need subquestions to be either an array or an object. We bounced around the idea of trying to allow QuestionComponent to handle both but ultimately decided that it made things too murky.

jgillis1 (approved):

PR #413, opened 2020-08-25
Generic section components that use path info

tadhg-ohiggins:
What section am I?
2B, perhaps; maybe not—
Check the URL.

Make sections generic components that look up specific sections based on the path.

### DIFF
--- a/frontend/react/src/store/formData.js
+++ b/frontend/react/src/store/formData.js
@@ -36,6 +36,17 @@ export const extractSectionOrdinalFromJPExpr = (jpexpr) => {
   return extractSectionOrdinalFromId(id);
 };
 
+/**
+ * @param int year: full four-digit year.
+ * @param int sectionOrdinal: just the digit; we don't expect it to have a leading zero here.
+ * @param string subsectionMarker: a–z or aa–zz. Should be lowercase by the time it gets here.
+ * @return string e.g. 2020-01-a.
+ */
+export const constructIdFromYearSectionAndSubsection= (year, sectionOrdinal, subsectionMarker) => {
+    const sectionChunk = sectionOrdinal.toString().padStart(2, "0");
+    return [year, sectionChunk, subsectionMarker].join("-")
+}
+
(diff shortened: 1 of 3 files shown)


frontend/react/src/store/formData.js
```diff
@@ -36,6 +36,17 @@ export const extractSectionOrdinalFromJPExpr = (jpexpr) => {
...
+ * @param int year: full four-digit year.
+ * @param int sectionOrdinal: just the digit; we don't expect it to have a leading zero here.
+ * @param string subsectionMarker: a–z or aa–zz. Should be lowercase by the time it gets here.
+ * @return string e.g. 2020-01-a.
+ */
+export const constructIdFromYearSectionAndSubsection= (year, sectionOrdinal, subsectionMarker) => {
+    const sectionChunk = sectionOrdinal.toString().padStart(2, "0");
+    return [year, sectionChunk, subsectionMarker].join("-")
+}
+
```
mgwalker, line 49: This and several other functions in `formData.js` that don't deal directly with redux state should probably be moved elsewhere. They're not strictly related to the data store. I think @alewoo28 is working on a better file organization structure than what I roughed in, so maybe after that's complete and merged, we can come back and move these general utilities somewhere else?

mgwalker (approved):
This looks good! Just tried it out and was delighted to see `localhost:81/sections/2020/0` rendered just like it ought. This seems like a nice stride towards fully dynamic rendering. Coupled with the work in #407 around questions and form elements, it seems to be coming together quickly.

Before we go any further down the path of everything having a `Data` prop, can we stop for a second and work on building an appropriate set of state selectors that provide only the necessary data instead of big state objects? In this case, for example, all the `Section` component seems needs is a list of subsection IDs. The `Subsection` component looks like it would need the subsection title and list of part IDs. And so on.

The `Data` prop is becoming an anti-pattern at this point, and if we keep perpetuating it, it's going to be harder to take back out. Should be fairly easy to do now, though, and sets us up to move quicker.

PR #414, opened 2020-08-26
Just Say No to Big Data

tadhg-ohiggins:
No to big Data
We don't need all the details.
We are nimble now.

Following @mgwalker 's comment (https://github.com/18F/cms-carts-seds/pull/413#pullrequestreview-474953228), pass only stripped-down data trees in mapStateToProps, rather than the full data trees we were previously passing.

### DIFF
--- a/frontend/react/src/store/formData.js
+++ b/frontend/react/src/store/formData.js
@@ -1,5 +1,6 @@
 import { LOAD_SECTIONS, QUESTION_ANSWERED } from "../actions/initial";
 import jsonpath from "jsonpath";
+import * as _ from "underscore";
 
 const initialState = [ ];
 
@@ -20,7 +21,7 @@ export default (sdata = initialState, action) => {
 export const selectSectionByOrdinal = (state, ordinal) => {
   const section = state.formData.filter(c => c.contents.section.ordinal === ordinal);
   if(section.length > 0) {
-    return section[0].contents;
+    return section[0].contents.section;
   }
   return null;
 }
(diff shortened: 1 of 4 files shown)


frontend/react/src/store/formData.js
```diff
@@ -20,7 +21,7 @@ export default (sdata = initialState, action) => {
 export const selectSectionByOrdinal = (state, ordinal) => {
   const section = state.formData.filter(c => c.contents.section.ordinal === ordinal);
   if(section.length > 0) {
-    return section[0].contents;
+    return section[0].contents.section;
```
tadhg-ohiggins, line 24: I changed this so that we no longer have to use it to get at the section everywhere else, but if anything outside the code here was already using it, it will now break as the `section` property it was expecting will be absent.

tadhg-ohiggins:
@alewoo28 this is probably of interest to you too, not least because of https://github.com/18F/cms-carts-seds/pull/414/files#r476914584.

frontend/react/src/store/formData.js
cannandev: I'm hitting `http://localhost:81/sections/2020/3/c` and seeing all section 3 data instead of the subsection. Am I missing something?

cannandev (approved):
My previous comment isn't related to this commit. Approving.

frontend/react/src/store/formData.js
alewoo28: OK, thanks for the headsup! It will be nice to skip that step of calling .section

PR #417, opened 2020-08-26
Feature/input grid api

cannandev:
This is implmented on 3C, and should work on 3G once @tadhg-ohiggins updates the API. Can you or @mgwalker walk me through using jsonpath? I feel like I shouldn't have as many maps as I do.
Also, I haven't figured out how to render the ordinal-subsection. It's displaying all of section 3.


jgillis1:
@cannandev If you pull feature/api_section3, we have it set up to go down one more level to get the subsection. Look at 3F.

PR #425, opened 2020-08-26
Hook up section 3F text fields to write to redux

mgwalker:
Just what it says on the tin

### DIFF
--- a/frontend/react/src/components/fields/QuestionComponent.js
+++ b/frontend/react/src/components/fields/QuestionComponent.js
@@ -1,9 +1,11 @@
 import React, { Component } from "react";
+import { connect } from "react-redux";
 import CMSChoice from "./CMSChoice";
 import { TextField, ChoiceList } from "@cmsgov/design-system-core";
 
 import DateRange from "../layout/DateRange";
 import CMSRanges from "./CMSRanges";
+import { setAnswerEntry } from "../../actions/initial";
 
 class QuestionComponent extends Component {
   constructor(props) {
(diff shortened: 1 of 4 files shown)


alewoo28 (commented):
Hey Greg. 

So on my branch, by changing the values of all of the questions type textfields to either question.answer.entry or " ", it breaks the ability to write anything in those text fields. Is that not a problem you're having in your version? 

It also looks like we've deleted this.props.sectionContext from handleChange and instead using  setAnswerEntry. I think that will work for textfields but not for anything that needs to be validated or have the data type altered. 

So we either have to have two separate actions-- 
-one that takes fragmentId & an event object 
and 
-one that takes a fragmentId and a formatted payload (like dates, ranges, emails or numbers that need validating, etc) as these things will have already been extracted from the event object to have been formatted-- so setAnswerEntry will fail because there is no target.value 

OR we have to artificially hand everything down through a fake object (in QuestionComponent) with a value nested in a target. ie: someFakeEvent = {
  target: {
    value: 'actual payload'
  }
}

Whatever we do-- I think we should get on the same page-- 1, so we're not doing double work (again), and 2, so that we proceed in the same format. Let me know what you think

alewoo28:
Also, I think its a better approach to send input straight from QuestionComponent. 
I was working on a different approach, but I can start implementing that instead now.

mgwalker:
- In section 3F, only the very last field (multiline text box for "7. Anything else you’d like to add that wasn’t already covered?") should be affected. That one still works on my branch. As long as the value is set to `question.answer.entry || ""` and the change handler is set to `this.handleChange`, other texty fields *should* work fine, but those are famous last words. (The `|| ""` is because if `question.answer.entry` becomes `null`, then React will break the control binding and then start complaining.) But you definitely need both pieces, the value and change handle binding.

- I took out the `sectionContext` part because when I traced it back to `Section3F`, it wasn't actually doing anything. `Section3F` had component state that was only ever written to, never read, so I chucked it because it looked like dead code. If there's a plan for that code, I'll gladly put it back! 🙂

- My assumption had been that each input type (for example, an email field) would handle its own validation and decide when/whether to pass changes along to redux. I'm reading that sentence back to myself right now and I realize it's not at all intuitive.

- I think we should change the `setAnswerEntry` action to take an answer value rather than an event object, because passing around an event object feels cumbersome. I mentioned that to @tadhg-ohiggins yesterday and I think he was also on board with doing that. I agree that wrapping up a dummy object just to pass to that object is not the best. 😬 

I **100% agree**, let's get synced up!

alewoo28:
Okay, I'll fiddle with the multiline text field then. It wont let me input but maybe im doing something wrong. I'll report back. 

Some of the validations might by trickier than that, but even still I think it can be done in QC right after the validations are complete. That makes sense to me. Anywhere where we were previously sending data (validated & properly formatted) to the parent component via this.props.sectionContext, instead we can chuck it straight to setAnswer to update the redux store. Fantastic!

alewoo28 (approved):

PR #426, opened 2020-08-26
Check if the fragment answer exists before getting its properties.

cannandev:


mgwalker (approved):
Approved, merge away when the `winnowProperties` issue is resolved!

PR #428, opened 2020-08-26
Idempotent command for loading basic fixtures into Django

tadhg-ohiggins:
Doing it again
Shouldn't change the outcome here.
Don't repeat yourself.

New Django management command `idempotent_fixtures` that only runs `loaddata` on a fixture if a matching document isn't already in the database.


mgwalker (approved):
Makes sense.

sure why not

PR #429, opened 2020-08-26
Add Django setup steps to entrypoint.sh

tadhg-ohiggins:
Django needs info.
Not what it knows already.
Only the new stuff.

Add a bunch of Django commands that make migrations, migrate, and load fixtures. These won't do anything if they've already been covered by prior deploys.

@dwhitecl This is the change we talked about earlier.

### DIFF
--- a/frontend/api_postgres/entrypoint.sh
+++ b/frontend/api_postgres/entrypoint.sh
@@ -11,4 +11,6 @@ then
     echo "PostgreSQL started"
 fi
 
+docker-compose run api_postgres sh -c "python manage.py makemigrations && python manage.py migrate && python manage.py idempotent_fixtures"
+
 exec "$@"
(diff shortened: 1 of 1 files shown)


frontend/api_postgres/entrypoint.sh
```diff
@@ -11,4 +11,6 @@ then
     echo "PostgreSQL started"
 fi
 
+docker-compose run api_postgres sh -c "python manage.py makemigrations && python manage.py migrate && python manage.py idempotent_fixtures"
```
mgwalker, line 14: I think this is being run inside the container, so you shouldn't need the `docker-compose` part.

```suggestion
python manage.py makemigrations
python manage.py migrate && python manage.py idempotent_fixtures
```

I'd favor double-checking with @dwhitecl to make sure that's correct.

mgwalker (approved):

PR #431, opened 2020-08-26
Make setAnswerEntry handle things that don't have .target.value

tadhg-ohiggins:
Stricter is better
But too many things will break
So, be tolerant.

Eventually we should accept only the unwrapped value; my putting it in an event-like object was a mistake. However, right now that would break a bunch of things, so this will put off the breakage while also allowing unwrapped values to work.

### DIFF
--- a/frontend/react/src/actions/initial.js
+++ b/frontend/react/src/actions/initial.js
@@ -13,10 +13,11 @@ export const loadSections = () => {
 };
 
 // Move this to where actions should go when we know where that is.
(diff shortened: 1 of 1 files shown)


tadhg-ohiggins:
@alewoo28 This should help with the concerns about needing to wrap things for setAnswerEntry from https://github.com/18F/cms-carts-seds/pull/425#pullrequestreview-475865425

mgwalker (approved):
good enough

...for now, though how would you feel about creating a tech debt issue to eventually come back and change this to only accepting a value?

PR #442, opened 2020-08-27
Improve lookup performance

tadhg-ohiggins:
JSON Path is slow
When we search through all the things.
Ask it to do less.

The major part of this PR is in switching to finding elements in the data structure using the knowledge we have of how it's constructed, rather than asking JSONPath to look through absolutely everything in it. The main chunk of code for this is `selectFragment` in formData.js.

This PR also contains working functional approaches to radio buttons and text boxes that communicate with the Redux store, although they don't yet support child questions or conditionals.

### DIFF
--- a/frontend/react/src/components/sections/basicinfoapi/QuestionLike.js
+++ b/frontend/react/src/components/sections/basicinfoapi/QuestionLike.js
@@ -1,6 +1,6 @@
 import React from "react";
 import { connect } from "react-redux";
-import { selectFragmentByJsonPath, winnowProperties } from "../../../store/formData";
+import { selectFragment, winnowProperties } from "../../../store/formData";
 import { setAnswerEntry } from "../../../actions/initial.js";
 import { ChoiceList, TextField } from "@cmsgov/design-system-core";
 import { _ } from "underscore";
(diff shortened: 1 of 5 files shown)


frontend/react/src/components/sections/basicinfoapi/QuestionLike.js
```diff
@@ -1,11 +1,11 @@
 import React from "react";
 import { connect } from "react-redux";
 import {
-  selectFragmentByJsonPath,
+  selectFragment,
   winnowProperties,
 } from "../../../store/formData";
 import { setAnswerEntry } from "../../../actions/initial.js";
-import { Choice, TextField } from "@cmsgov/design-system-core";
+import { ChoiceList, TextField } from "@cmsgov/design-system-core";
```
cannandev, line 8: @tadhg-ohiggins I'm approving... But I'd like to understand why we're going back to `ChoiceList`?
`Choice` supports `checkedChildren` for conditional questions.

cannandev (approved):

frontend/react/src/components/sections/basicinfoapi/QuestionLike.js
tadhg-ohiggins: We can go back! I was having a lot of trouble with `Choice` and some Redux interactions, but can move the current questions back if we need to. We should talk this over.

PR #448, opened 2020-08-27
Make the API URL dynamically settable, so it can work in all environments

mgwalker:
- update the Axios calls to use `process.env.REACT_APP_API_URL` for the API URL base
- update the dev Docker configuration to set `REACT_APP_API_URL`
- update a Terraform template to set `REACT_APP_API_URL` in the ui container

**Note:** I used the name `REACT_APP_API_URL` because all environment variables that start with `REACT_APP_` are automatically injected by `create-react-app`, so by using that prefix, we get the environment support for free.

### DIFF
--- a/frontend/docker-compose.dev.yml
+++ b/frontend/docker-compose.dev.yml
@@ -20,8 +20,9 @@ services:
     networks:
       - data_net
     environment:
-      - API_URL=http://localhost:8000 # This is how a browser would call the api
       - CHOKIDAR_USEPOLLING=true
+      - CHOKIDAR_INTERVAL=3000
+      - REACT_APP_API_URL=http://localhost:8000
     stdin_open: true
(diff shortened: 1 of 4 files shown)


frontend/docker-compose.dev.yml
```diff
@@ -20,8 +20,9 @@ services:
     networks:
       - data_net
     environment:
-      - API_URL=http://localhost:8000 # This is how a browser would call the api
       - CHOKIDAR_USEPOLLING=true
+      - CHOKIDAR_INTERVAL=3000
```
mgwalker, line 24: This slows down the filesystem polling interval so that running Docker hopefully no longer sounds like a rocket launching for the moon, even though lunar missions are cool.

dwhitestratiform (approved):

PR #453, opened 2020-08-27
Make the deployment script run the right Django application.

tadhg-ohiggins:
Not hello_django
But rather run the carts app.
It is for the best.

Switch the deployment docker-compose.yml to invoke the carts Django app.


cannandev (approved):

mgwalker (approved):
a plus

Issue #449, opened 2020-08-27
Make side navigation dynamic


mgwalker:
Use the section/subsection info from the API to dynamically build the sidebar and its links

PR #461, opened 2020-08-28
Make the sidebar dynamic

mgwalker:
- hooks up the sidebar to the list of sections and subsections that come from the API
- adds a section title for section 3
- addresses #449 

screenshot

For the sections that won't be using the dynamic section component, we can handle those by explicitly adding those routes to the route handler. Shouldn't be a biggie.


jgillis1 (approved):
This is very cool. Nicely done! Also nice touch on the notes about what is missing JSON.

Issue #446, opened 2020-08-27
Connect generic section to question component


mgwalker:
Acceptance criteria:
- [ ] Generic section components (maybe actually `Part.js`) uses the `QuestionComponent` to render questions dynamically
  - [ ] section component passes appropriate JSONpath info down to question component
- [ ] `QuestionComponent` pulls question info from redux

PR #464, opened 2020-08-28
WIP: Section 3F & 3E Reading & writing 

alewoo28:
**WIP**

Done: Sections 3F & 3E Read from the api fine and most questions write to redux just fine. 

**Problem:**
Nested fieldset questions. When nested fieldset questions take in input, theres a problem with the selectQuestion function. 
In the JSON look at question "2020-03-e-01-12".  Then look at the selectQuestion function in store/selectors
When jp is changed to  const jp = `$..[*].contents.section.subsections[*].parts[*].questions[*].questions[?(@.id=='${id}')]`;
Then there is no problem with editing that nested question (but then there IS a problem with editing questions that arent double nested)

Maybe this is an easy fix? 

I'm making a PR for this, but it is a WIP, very nearly done!


tadhg-ohiggins:
@alewoo28 I haven't had time to investigate this in detail, but does this work if you use `selectFragment` instead of `selectQuestion`?

alewoo28:
Hey Tadgh, Claire and Greg

Ty for the suggestion. So i fiddled with this and i'm coming up to a different variation of the same problem. 

In the case of nested questions (ie: fieldsets, like 2020-03-e-01-12, 13 & 14) in the reducer:
When we use selectQuestion, const fragment is assigned to null. 
When we use selectFragment, const fragment is assigned to the entire fieldset array. 

So in both cases setting fragment.answer.entry breaks. 

selectFragment at least is giving us the array of questions. 

Do y'all have any suggestions on where/when the best place would be to parse through that?

One solution we had was to duplicate the selectFragment function and in 
 this selectQuestionFromFragment function, before it returns, check to see if it contains a question array and parse through that for the specific question we're looking for. 

I didnt want to edit selectFragment because I see its being used in other places and dont want to break any functionality elsewhere. 

Let me know what you think or if y'all have another idea of how we can do that

mgwalker:
I think if a question is not found, we should immediately check the fieldsets for that question. I.e., update `selectionQuestion` to do a second query if necessary:

```js
export const selectQuestion = (state, id) => {
  const jp = `$..[*].contents.section.subsections[*].parts[*].questions[?(@.id=='${id}')]`;
  const questions = jsonpath.query(state, jp);
  if (questions.length) {
    return questions[0];
  }

  const jpFieldset = `$..[*].contents.section.subsections[*].parts[*].questions[?(@.type=='fieldset')].questions[?(@.id=='${id}')]`;
  const fieldsetQuestions = jsonpath.query(state, jpFieldset);
  if (fieldsetQuestions.length) {
    return fieldsetQuestions[0];
  }

  return null;
};
```

alewoo28:
Okay cool, thanks Greg

will the new jpFieldset be able to account for doubly nested fieldsets? like fieldsetQuestions/fieldsetQuestions/question in "2020-03-b-01-03"

alewoo28:
This PR is ready. All question types are reading and writing, even questions nested in fieldsets. This leaves 3E & 3F in pretty good shape. 
Conditional logic needs work & the file upload Component needs to be made, but this PR doesnt need to wait for that.

tadhg-ohiggins:
Is the version of the Section 3 JSON included in this PR relevant? It looks as if it's whitespace changes. If there are no intentional changes, please remove it from the PR; if there are intentional changes, please revert to the original whitespace and then add just those changes so that they're clearly visible.

frontend/react/src/components/fields/QuestionComponent.js
```diff
@@ -20,6 +20,7 @@ class QuestionComponent extends Component {
     this.handleIntegerChange = this.handleIntegerChange.bind(this);
```
mgwalker, line 20: Ideally I'd like to see this broken into composable components instead of one big monolithic thing. There is a _*lot*_ of logic built into this single component that really ought to exist in more. For example, there could be a `<PhoneNumberQuestion>` that handles its own validation. Then here in `QuestionComponent`, it'd render a `PhoneNumberQuestion` instead of doing all the logic itself. Compartmentalization and the single-responsibility principle!

As it is, the `QuestionComponent` is responsible for a bunch of things, which makes it harder to maintain and a more likely source of future bugs.

I don't think we should hold this PR up over it with a looming deadline, but I do think we should open an issue to make sure we eventually come back to this and break it up.

frontend/react/src/store/saveMiddleware.js
```diff
@@ -1,6 +1,7 @@
 import axios from "axios";
 
 import { QUESTION_ANSWERED } from "../actions/initial";
+import { QUESTION_FORMATTED } from "../actions/initial";
```
mgwalker, line 4:
```suggestion
```

Not used

frontend/react/src/store/selectors.js
```diff
@@ -1,10 +1,19 @@
...
+  if (questions.length) {
     return questions[0];
   }
+
+  // const jpFieldset = `$..[*].contents.section.subsections[*].parts[*].questions[?(@.type=='fieldset')].questions[?(@.id=='${id}')]`;
+  // const fieldsetQuestions = jsonpath.query(state, jpFieldset);
+  // if (fieldsetQuestions.length) {
+  //   return fieldsetQuestions[0];
+  // }
+
```
mgwalker, line 15:
```suggestion
```

delete commented lines

frontend/react/src/store/selectors.js
```diff
@@ -1,10 +1,19 @@
...
   return null;
-}
+};
+
+// const jp = `$..[*].contents.section.subsections[*].parts[*].questions[*].questions[?(@.id=='${id}')]`;
```
mgwalker, line 19:
```suggestion
```

delete commented lines

mgwalker (changes requested):

frontend/react/src/components/sections/section3Eapi/Section3E.js
```diff
@@ -0,0 +1,91 @@
+import React, { Component } from "react";
```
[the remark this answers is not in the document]
jgillis1: Since we had already been building these out prior to the generic section component being circulated, it was decided we would continue to use these pages. Also, when I look at the demo site, the generic section component does not render 3E or 3F (though I have not explored why) so it seems like these are needed.

frontend/react/src/components/sections/section3Eapi/Section3E.js
```diff
@@ -0,0 +1,91 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  render() {
+    // this.props.Data includes all of section 3
+    // This variable narrows it down to a subsection
+    const subsectionData = this.props.Data
+      ? this.props.Data.subsections[4] // 3E JSON Data
```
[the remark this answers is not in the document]
jgillis1: Understood

frontend/react/src/components/sections/section3Eapi/Section3E.js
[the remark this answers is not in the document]
mgwalker: I realized late last night that hooking up the generic component to the question component was a separate task in #446, so my comment here was misplaced! Sorry for throwing confusion into the mix.

if @andrewadcock hasn't started on it yet, I could tackle it a little later today. There's a minor bug in one piece of the generic section component that's making a couple of sections blow up, but I have a fix locally, and then hooking up the question component should be fairly straightforward after that.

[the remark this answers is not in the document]
jgillis1: Assigned to you!

frontend/react/src/components/sections/section3Eapi/backend-json-section-3.json
```diff
@@ -0,0 +1,1523 @@
+{
```
mgwalker, line 1: Can this file also be deleted?

mgwalker (approved):

frontend/react/src/components/sections/section3Eapi/backend-json-section-3.json
alewoo28: can do

alewoo28:
All the new merge conflicts are taken care of

PR #485, opened 2020-09-01
Hook up the generic section to the question component

mgwalker:
- moves the generic Section/Subsection/Part components from `basicinfoapi` to `components/layout`
- tweaks the behavior of the generic Part component to pull titles/text in the selector, and inline some stuff; also pass a part ID on to the QuestionComponent
- updates the QuestionComponent to fetch a list of questions based on a part ID
- minor fix to the `shouldDisplay` helper to handle questions that don't have a context (assumes they should be shown)
- addresses #446


frontend/react/src/components/Utils/helperFunctions.js
```diff
@@ -7,6 +7,9 @@ const sliceId = (id) => {
 };
 
 const shouldDisplay = (parentValue, context) => {
+  if(!context) {
+    return true;
+  }
```
mgwalker, line 12: If someone could verify this behavior, that'd be dandy. It looks like there are some questions that get passed into the `CMSChoice` component that do *not* have a `context` property. In that case, I assumed that the question should always be shown and bailed out early. Otherwise there's a null reference error trying to check the condition.

frontend/react/src/components/fields/QuestionComponent.js
```diff
@@ -448,8 +450,12 @@ class QuestionComponent extends Component {
 // "daterange"             <<831
 // "mailing_address"         <<831
 
+const mapStateToProps = (state, { data, partId }) => ({
+  data: data || selectQuestionsForPart(state, partId),
+});
+
```
mgwalker, line 456: The QuestionComponent can work one of two ways:

1. you pass in a `partId` prop, and that will be used to fetch a list of questions
2. you pass in a `data` prop with a list of questions to be rendered, and it will use that; or

The first case is used by the section to get things started, and the second case is used by the QuestionComponent itself to render sub-questions/fieldsets/etc.

frontend/react/src/components/Utils/helperFunctions.js
alewoo28: Hey greg, this makes sense to me. For radios & checkbox children that may not have contexts

frontend/react/src/components/fields/QuestionComponent.js
alewoo28: Wont having a data prop in mapStateToProps break the hard coded sections that are passing the questons down using data={part.questions} because there will be two different props.data 

Maybe the solution is to change the name of the Data prop in the hard coded sections?

alewoo28: I just traced through Sections.js > Subsection.js > Part.js > QuestionComponent and seeing how they each are being informed by redux instead of prop drilling is *100 emoji*. Its a fantastic pattern. 

My next thought is about nested components. When QC has children questions and is called recursively, recursive children wont have access to redux at all (except through props).  Do you forsee this being a problem?

mgwalker: It will, yes, but the intent of this PR is to switch to the dynamic section component and not use the hard coded sections at all anymore.

alewoo28: Alright cool! That makes sense. I guess it would be identical data anyway. I think for some of them we will still have to not send the questions array straight from redux (ie: 2b), but I guess we cross that bridge when we get there. I think overall this pattern is awesome. 

I also answered my own question about nested components. Children not accessing redux is a good thing anyway so that in the case of data={question.questions} for children, they should be reading from that data prop and not redux anyway! fantastic

alewoo28 (approved):

PR #494, opened 2020-09-02
Change checklist/radio options from unordered objects to ordered arrays

### LINKED ISSUE #474: Options Obj. Doesn't Present in Order
The Options Object doesn't present in proper order.

mgwalker:
The current `options` property on `question.answer` is an object. In Javascript, object properties are unordered, so we can't guarantee that the order we put things in is the order that they'll show up on the screen. Here's what it looks like now:

```json
{
  ...
  "options": {
    "User-facing text": "data value",
    "Another option": "data 2",
    "One more": "data 3"
  }
}
```

To guarantee order, we're switching the `options` property to an array. Arrays have guaranteed order, so now we have this:

```json
{
  ...
  "options": [
    { "label": "User-facing text", "value": "data value" },
    { "label": "Another option", "value": "data 2" },
    { "label": "One more", "value": "data 3" },
  ]
}
```

## Key things to review
- the [changes to the JSON schema](https://github.com/18F/cms-carts-seds/pull/494/files#diff-c2ef63b08e3cde05b79b5f9f48e37910) - all the JSON files validate, so if the schema is good, then so are the JSON files
- the [frontend changes](https://github.com/18F/cms-carts-seds/pull/494/files#diff-e94bf0733438bdcdaebae4c2f7046400) (starting at the link and then going down) - I updated the question component as well as all the statically-rendered sections, just in case

---

- updates all the JSON and fixtures
- updates the temporary hardcoded JSON that's serving in place of the API
- updates the frontend components that use `question.answer.options`
- closes #474


docs/section-schemas/2020-ak-section-3.json
```diff
@@ -773,10 +996,9 @@
                         "questions": [
                           {
                             "comment": "This is the first real question so far in this part…",
-                            "type": "question",
+                            "type": "integer",
                             "id": "2020-03-c-04-01-a",
                             "label": "Total for all ages",
-                            "type": "integer",
```
mgwalker, line 779: A lot of these questions have two `type` properties. The second one takes precedence, so the JSON serializer removed the first one when outputting it back out. The upshot is that the JSON is now internally valid. 😛

tadhg-ohiggins: Thanks for catching that!

tadhg-ohiggins (approved):

alewoo28 (approved):
I think an array makes more sense than an object. Looks great! 

I know that @TGriesemer is working on section 2b though, so we should give him a heads up about changes in
frontend/react/src/components/sections/section2api/ObjectiveAndGoals.js and frontend/react/src/components/sections/section2api/objectives/Objective2bApi.js

But it should be all good, i dont think it should mess with the functionality

mgwalker:
Is section 2B going to continue to be a static section rather than dynamic? 🤔

alewoo28:
I think at least somewhat static? Its building quasi-dynamically from the API like the obsolete 3F/3E did, but im not sure that the dynamic section generator could grapple with 2B. It has a bunch of nested 'repeatables'. It doesnt just populate question.answer.entry, it also has to build new repeatables. Tim could explain whats going on there better than I can, i dont want to mispeak on this.

PR #498, opened 2020-09-03
ENABLE-198

sethsacher:
Add a configuration for the number of container instances deployed per service. Default is 1 per service.


sethsacher:
Replaced by https://github.com/18F/cms-carts-seds/pull/500

PR #502, opened 2020-09-03
Add a GitHub action to validate JSON files against the schema

mgwalker:


tadhg-ohiggins (commented):
Looks good to me, but I made some changes of my own that you should probably look over.

tadhg-ohiggins (approved):

Issue #497, opened 2020-09-03
Conditional questions need some tweaking


mgwalker:
For example, in section 1, question 2 is the parent for three other questions.

![Screen Shot 2020-09-03 at 9 48 38 AM](https://user-images.githubusercontent.com/1775733/92130900-e33fb000-edca-11ea-9519-890399fd2b44.png)

---

If the answer to question 2 is yes, then questions 2a and 2b appear:

![Screen Shot 2020-09-03 at 9 48 47 AM](https://user-images.githubusercontent.com/1775733/92130991-09655000-edcb-11ea-89bd-aa967f5ba8d8.png)

---

If the answer to question 2a is no, then question 2c appears:

![Screen Shot 2020-09-03 at 9 48 57 AM](https://user-images.githubusercontent.com/1775733/92131040-18e49900-edcb-11ea-90d8-7629445e89a1.png)

---

All good so far, but if you then change question 2 to no, question 2c does not go away:

![Screen Shot 2020-09-03 at 9 49 06 AM](https://user-images.githubusercontent.com/1775733/92131092-27cb4b80-edcb-11ea-8468-2923ef5ff5f4.png)

---

The problem is that the answer to 2a is still "no," even though it is hidden. There are two obvious paths forward here:

1. Make the conditional logic recursive so that when we check if 2c should be visible, we also check if the question it depends on is visible. **I don't recommend this approach.** It is complicated, in the best case, and dangerous in the worst case. There is likely to be a pretty significant performance hit (users would see a delay between when they click on things and when the UI updates), and recursion always carries risks of bugs caused by not stopping at the right time.

2. Update the JSON to wrap questions 2a, 2b, and 2c in a fieldset and make the whole fieldset dependent on the answer in question 2. This seems safer and more maintainable long-term. We would probably need to go through the whole app to find all the places where this is happening and then update them all at once. I _think_ if we can get a list of questions that need to be addressed, we may be able to write a script to go fix them all in one go.

tadhg-ohiggins:
Greg's second suggestion is the one I'd recommend. The current JSON doesn't quite capture enough of the conditional logic this particular set of questions needs.

cjhskippy:
@erinzstrenio - perhaps your extraordinary powers of organization might help us catalog all of the questions to which this scenario applies?

erinzstrenio:
It should only apply to a few questions in Section 1 (of the whole report): 
- Part 1 Q2, Q3 
- Part 2 Q2, Q3

PR #504, opened 2020-09-04
Fix question relationship in section 1a, question 2

mgwalker:
- addresses #497


mgwalker:
@tadhg-ohiggins This one is good to go now. It also includes a couple of fixes to the Github action for validations, since the pathing was wrong on it. I never know where the heck anything is going to run...

tadhg-ohiggins (approved):

PR #519, opened 2020-09-08
redo SEDS SQL so it can (hopefully) be pulled into Section 2a

jgillis1:
Three new SQL files that I think will query the SEDS data in a more usable fashion

- SCHIP_Q4_Line7Total for Separate CHIP

- MCHIP_Q4_Line7Total for Medicaid expansion CHIP

- All_CHIP_Q4_Line7Total for showing Separate CHIP and Medicaid expansion CHIP results combined

Results from ALL_CHIP_Q4_Line7Total looks like this (note dummy data is shown below not actuals):
image

### DIFF
--- a/docs/All_CHIP_Q4_Line7Total.sql
+++ b/docs/All_CHIP_Q4_Line7Total.sql
@@ -1,0 +1,74 @@
+Declare @currentYear as integer
+
+Set @currentYear = 2016;
+
+with MCHIP as (
+	Select StateCode
+		  ,SubmissionDate
+		  ,sum(LineTotal) as Line7Total
+	from (
+		SELECT [StateCode]
+				,[SubmissionDate]
+				,[AgeRange]
+				,[LineNumber]
+				,sum(IsNull([Column1],0) + IsNull([Column2],0) + IsNull([Column3],0) + IsNull([Column4],0) + IsNull([Column5],0)) as LineTotal
+			FROM [SCHIP].[dbo].[tbl64_21E]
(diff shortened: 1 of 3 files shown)


kaitlin:
@jgillis1 where is this sql supposed to be executed? Is this something you run locally to pull data from SEDS?

jgillis1:
@kaitlin We haven't discussed how to implement this yet. To build it, yes, I was running against my local SEDS DB. Ultimately, the query itself will be used to pull the data from the AWS replicated SEDS database that is needed to populate the tables in Section 2a. Since the structure of the SEDS database is highly normalized, I put the queries together to help ensure we pull the correct data. I have two thoughts on implementation:

1. Set up a job on the server side to copy the data from the SEDS database into the appropriate location in our CARTS database. The job would run after the SEDS replication is complete for the day.

2. We can use the query in an API call to grab the data when Section 2a is rendered. 

Option 1 seems like the easier route to take since the SEDS data will only be replicated to AWS once a day at this point. Once the SEDS redesign is complete, we would need to look at changing this process up to get a live feed from the new SEDS database. At that point we would want go to with Option 2.

mblevins-collabralink (approved):
I forgot how, um, "creative" the table names were in the old system. The syntax looks fine, though of course we can only test locally. Neither of these results sets should be very large, so the union isn't scary.

kaitlin:
I'll merge then, since I think realistically you two know this data the best. I think at sprint plan we should have a conversation about the data migration in general and ensure we've got tickets for everything.

cc @StephanieCoppel @cjhskippy

PR #520, opened 2020-09-08
Run QC through prettier.

cannandev:
Runs QC through `prettier` until we add it as a github action. Make sure you have `prettier` enabled locally in your code editor.


jgillis1 (approved):

jgillis1:
This is interesting because we do all use prettier.

Issue #480, opened 2020-09-01
Re-enable authentication


cjhskippy:
In #479, we circumvented the authentication mechanism for the purpose of allowing the project team to work and test from a variety of user roles.  Prior to launch or accepting sensitive information, we need to re-enable those authentication controls.

Note: At the moment, some of the devs do not have working Okta credentials.  In order to ensure that development can continue, the bypass will continue to work in dev + local environments.  This can be removed once all devs have working Okta authentication.

AC:
- [ ] Make auth circumvention controlled by an environment variable
- [ ] Set environment variables as follows:
  - [ ] OFF for Prod
  - [ ] OFF for Val
  - [ ] OFF for Test
  - [ ] ON for Dev

StephanieCoppel:
completed.

Issue #479, opened 2020-09-01
Temporarily circumvent authentication


cjhskippy:
Since the project team does not all have valid EUA accounts, we need to temporarily circumvent the login process so that we can continue development while enabling features related to login, user groups, and permissions.

Issue #392, opened 2020-08-20
User Authentication 


cjhskippy:
Collection of stories related to user authentication and activities permitted for each of the user roles.

Necessary for states to begin their 2020 CARTS report: 
- [x] Temporarily disable authentication ( #479 )
- [ ] User Roles in database ( #206 )
- [x] Connect React -> Django ( #481 )
- [x] Login page ( #300 )
- [ ] Re-enable authentication ( #480 )

PR #526, opened 2020-09-09
Okta auth with fake user data bypass

tadhg-ohiggins:
Enable Okta—
But leave another way in
With parameters.

Merging this will wrap the application in Okta authentication, going through https://test.idp.idm.cms.gov/ and requiring CMS EUA credentials. Even on localhost. However, appending one of:

- `?dev=dev-ak` 
- `?dev=dev-az` 
- `?dev=dev-ma` 

to the URL will bypass this and provide access as a fake user associated with the corresponding state.

Currently CMS EUA authentication isn't working for me at all, even though it worked fine last week. That doesn't appear to be related to this PR, though.

Testing it locally before merging would probably be good…

This also adds support for the `context_data.show_if_state_program_type_in` logic at the Part level; this was basically necessary to add as a way of verifying that the different states' info was being picked up from the fake data.

Addresses #479, also moves #480 and #392 along. Probably makes #376 unnecessary.


frontend/react/src/components/layout/Header.js
```diff
@@ -55,7 +56,6 @@ class Header extends Component {
                       <a href="javascript:void(0)">Manage account</a>
                     </li>
                     <li className="logout">
-                      <a href="javascript:void(0)">Log out</a>
```
mgwalker, line 58: Does this need `<Logout />` added?

frontend/react/src/actions/initial.js
```diff
@@ -6,6 +6,7 @@ export const QUESTION_ANSWERED = "QUESTION ANSWERED";
 const temp__data = require("./initial.json");
 
 export const loadSections = () => {
+  console.log("loadSections");
```
mgwalker, line 9:
```suggestion
```

mgwalker (approved):
I don't love it, but it seems like a good iteration towards the goal. It's hard to love some of the incremental steps. 🤷‍♂️ 

I added a lot of suggestions for removing unused code/log statements and one question, but nothing worth holding this up for. I'm going to approve it and let you decide whether to merge now or clean anything up first.

frontend/react/src/components/layout/Header.js
tadhg-ohiggins: It needs that handled somehow; right now, if that component is rendered the whole page errors out if you're not authenticated via Okta. I don't think that should hold up the PR, though.

PR #525, opened 2020-09-08
Update Section 1 per DSCP feedback

mgwalker:
- adds a `hide_if_all` display conditional
- updates section 1 to use the `hide_if_all` conditional
- closes #415 

@erinzstrenio: At the bottom of section 1a, part 3, I added a new question. Eventually this question will be hidden if the answer to the previous 19 (❗ ) is no, but for your review, here's what it looks like now. I wasn't sure what the question text should be, so I'm asking. 😄

screenshot of new question

PR #529, opened 2020-09-09
Feature/conditional logic

alewoo28:
This is a WIP, either all of the conditional children show or none of them do. Its very close though 

**Branch:**
feature/conditional-logic
**Files:**

selectors.js (_selectQuestionsForPart and filterDisplay_) ~line 50
formData.js (_shouldDisplay_) ~line 259

**selectQuestionsForPart** is generating the array of questions to iterate through. Its supplying those to QuestionComponent in the mapStateToProps. We are filtering the data before it gets to QC so that questions that should not render, are not present in the data array.

**filterDisplay** is a function called to filter the question in the array and calls itself recursively when a question has subquestions.

**shouldDisplay** is a function that takes in a single question's question_context and the redux state. It returns a single boolean establishing if a question should display or not. It needs the entire redux store in order to run jsonpath.query


alewoo28:
This is ready for review now, no longer a WIP

alewoo28:
### More notes on the files changed: 

**frontend/react/src/components/Utils/helperFunctions.js:** 
The previous home of the shouldDisplay function. It was removed from this file. 

**frontend/react/src/components/Utils/questionUtils.js:**
Removed shouldDisplay import, not being used here anymore 

**frontend/react/src/components/fields/CMSChoice.js:**
Removed where CMSChoice was using shouldDisplay. That check is no longer being made in CMSChoice

**frontend/react/src/components/fields/QuestionComponent.js:**
changed formating (removed arrow function syntax from class method). but this will be removed anyway, the linter was just bugging me about it. 

**frontend/react/src/components/sections/section1api/Section1.js:**
these are linter changes. I didnt make any actual changes here. 

**frontend/react/src/components/sections/section1api/backend-json-section-1.json:**
Made this JSON file the same as the JSON in section schas 

**frontend/react/src/components/sections/section1api/questions/Questions1api.js:**
Linter changes

**frontend/react/src/store/formData.js:**
The shouldDisplay function was added to this file 

**frontend/react/src/store/selectors.js**
selectQuestionsForPart has been edited to filter questions (using filterDisplay)
filterDisplay fucntion added

frontend/react/src/store/formData.js
```diff
@@ -253,3 +253,38 @@ export const generateSubsectionLabel = (str) => {
...
+
+/**
+ * This function checks to see if a question should display based on an answer from a different question
+ * @function shouldDisplay
+ * @param {object} state - The application state from redux, the object required for jsonpath to query
+ * @param {object} context - the context_data from a question
+ * @returns {boolean} - determines if an element should be filtered out
+ */
+export const shouldDisplay = (state, context) => {
+  if (!context || !context.conditional_display) {
```
mgwalker, line 268: We'll need to extend this a bit in the future, because soon there will be questions that have `conditional_display` but do not have `hide_if` (we're adding a new condition type in #525). Nothing to change now, though.

mgwalker (approved):
Looks good to me!

Issue #505, opened 2020-09-04
Add Prev/Next Buttons to Section Component


jgillis1:
Prev/Next Buttons are visible at the bottom of each Section.

AC:

- [x] Next button is visible for all Sections that have a following Section (currently 0, 1, 2a, 2b, 3a-3f)
- [x] As new sections are added to the report, the previous and next buttons appear in the appropriate sections without introducing code changes.
- [x] Next button navigates to the next subsection.  If there is no following subsection, the button navigates to the next section
- [x] Previous button is visible for all Sections except Section 0
- [x] Previous button navigates to the prior subsection.  If there is no prior subsection, it navigates to the prior Section.
- [ ] Styling for the previous and next buttons should conform to designs in #113

andrewadcock:
@cjhskippy there is currently a button in the footer labeled "All changes auto-saved". Given that our saving strategy has changed, is this still needed? Or should we hook it to the current save functionality (not sure the feasibility of this)?

cjhskippy:
@andrewadcock Please ignore that "save" notification.  It's out of the scope of this ticket, and won't really be meaningful until we're saving to a database.  If it's in the way of the next / previous buttons, you're welcome to hide it.

cjhskippy:
I see a couple of potential issues with this - will connect with Stephanie to see if those should be broken out into new issues or if I should reopen this one.

- The "next" button on the "Basic Info" page does not go to Section 1.  Instead it tries to open a section "0a".
You can reproduce this by going to https://mdctcartsdev.cms.gov/sections/2020/00 and clicking "next".  The URL is changed to https://mdctcartsdev.cms.gov/sections/2020/00/a
- The next button is sometimes left in the "pressed" state (yellow background) even after a page loads.  You can reproduce this by going to https://mdctcartsdev.cms.gov/sections/2020/00 and clicking "next".

Issue #113, opened 2020-06-24
Design global form actions component 


austinhernandez:
The bottom of each page in the report will have the same actions displayed consistently across the application. The actions a user can take in each section are:

- [x] Go to the previous page
- [x] Go to the next page
- [x] Let a user save or understand that the document has been auto-saved
- [x] Let a user share the document in its current state with other members of their organization to fill out or view.
  - We were calling "export" "share" "download and print" but due to technical constraints is more accurately and realistically, "This will bring you to a page where you can print or save as PDF and share with others." This doesn't quite fit in a button or roll of the tongue so this section still needs work.

<img width="839" alt="Screen Shot 2020-06-24 at 7 07 25 PM" src="https://user-images.githubusercontent.com/23194720/85636250-f6f1cc00-b64d-11ea-8be7-001e754f3eaa.png">

### Dependency and blockers:
- [x] Define the process and technical limitations of the "export" functionality
- [x] Design a non-editable view/review view.
- [x] Design a `print.css` view

Note: 
- Very strong candidate for next round of usability testing

erinzstrenio:
@austinhernandez Looks great!

Is there a "save" button too or just the auto-save function? Curious since you noted the action to "Let a user save..."

austinhernandez:
@erinzstrenio document will auto-save, letting a user save was a fallback in case this was not technically feasible.

austinhernandez:
Notes from co-work: 6/25

- The actions aside from go to the previous page, go to the next page will be to generate a temporary URL to share (`share this page`) and open the print preview dialog box so they can print or save as a PDF (`print`). 
  - names subject to change. 

- `Share this page` Will generate a temporary URL that a user can copy and send (via email, chatter, gchat... etc.) to a colleague to fill out. It will share only that page/subsection with the data from the previous year available. only the to share and 
- Print: Will open the print preview page [essentially like pressing ctrl/cmd P]


- [ ] This ticket will just be the global footer design with auto-save, previous, next, print, and share this page buttons.
- Design review view is in a new ticket #118 
- Design the `print.css` view of the subsections is a new ticket #119

austinhernandez:
right?
<img width="816" alt="right" src="https://user-images.githubusercontent.com/23194720/85818362-7d85d680-b73e-11ea-8776-6cafae1df51b.png">

outside?
[...]

PR #530, opened 2020-09-09
Add dynamic form buttons (next, prev) to all sections

andrewadcock:
Adds dynamic next and previous buttons (#505). Also adds back the share and print buttons and design from #113.

`Share` button is still using dummy data (Covered in #403 )

`All changes auto-saved` button has been moved to a new component and styled.


frontend/react/src/components/layout/FormNavigation.js
```diff
@@ -2,50 +2,135 @@ import React, { Component } from "react";
...
+
           </div>
         </div>
       </section>
     );
   }
 }
 
-export default FormNavigation;
+const selectSectionsForNav = (state) => {
```
cannandev, line 100: Can we move this to the `selectors.js`?

andrewadcock: Successfully moved to selectors.js. Thanks!

frontend/react/src/components/layout/FormNavigation.js
```diff
@@ -2,50 +2,104 @@ import React, { Component } from "react";
...
+import { connect } from "react-redux";
+import { withRouter } from "react-router-dom";
+import { selectSectionsForNav } from "../../store/selectors";
+
+const idToUrl = (id) => `/sections/${id.replace(/-/g, "/")}`;
 
 class FormNavigation extends Component {
+
   render() {
+    const { location, sections } = this.props;
```
mgwalker, line 14:
```suggestion
    const { history, location, sections } = this.props;
```

Pull in the history object. See the later comment about why!

frontend/react/src/components/layout/FormNavigation.js
```diff
@@ -2,50 +2,104 @@ import React, { Component } from "react";
...
+      if (parentItems[i].url !== undefined) {
+        items.push(parentItems[i].url);
+      }
+
+      if (parentItems[i].items !== null) {
+        for (let j = 0; j < parentItems[i].items.length; j++) {
+          items.push(parentItems[i].items[j].url);
+        }
+      }
+    }
```
mgwalker, line 52: This seems like it's a lot of effort just to derive URLs from IDs. How about something simpler:

```js
const items = [];
sections.forEach(section => {
  section.subsections.forEach(subsection => {
    items.push(idToUrl(subsection.id));
  });
});
```

There might be a little more weirdness to account for sections with a single subsection (where the URL is something like `sections/2020/00/`, for example, but it doesn't make sense to me to have a pair of maps with a nested map when the final output is just these URLs.

frontend/react/src/components/layout/FormNavigation.js
```diff
@@ -2,50 +2,104 @@ import React, { Component } from "react";
...
           <div className="ds-l-col form-buttons">
-            {this.props.previousUrl ? (
+
+            {previousUrl ?
               <div className="form-button previous">
                 <Button
                   type="submit"
                   className="ds-c-button"
-                  href={this.props.previousUrl}
+                  href={previousUrl}
```
mgwalker, line 77:
```suggestion
                  onClick={() => { history.push(previousUrl); }}
```

Use the history object to push instead of using an `<a>` (that's how these buttons render if there's a `href` property attached). That way we don't hard-reload the site and instead just do an internal navigation.

frontend/react/src/components/layout/FormNavigation.js
```diff
@@ -2,50 +2,104 @@ import React, { Component } from "react";
...
-            {this.props.nextUrl ? (
+              </div> : null}
+
+            {nextUrl ?
               <div className="form-button next">
                 <Button
                   type="submit"
                   className="ds-c-button ds-c-button--primary"
-                  href={this.props.nextUrl}
+                  href={nextUrl}
```
mgwalker, line 88:
```suggestion
                  onClick={() => { history.push(nextUrl); }}
```

mgwalker (changes requested):
Only blocking change I see is that the browser hard reloads when you click the button, which means there are also a series of roundtrips to the API, reloading the state, etc. I've proposed 3 lines of changes that will use `history.push` instead for a soft navigation that should take care of it.

I had a couple of other questions/comments about how to clean the code up a little, but overall I think this is in good shape. 😄

frontend/react/src/components/fields/Autosave.js
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from "react";
```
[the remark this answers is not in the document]
andrewadcock: Got an answer: Keep it in a separate component and we will deal with it later.

mgwalker (approved):
We need to add an issue to scroll the window back to the top when the user clicks the nav buttons, and make sure we handle focus appropriately when that happens, but this PR is ready to go.

PR #541, opened 2020-09-10
Improve jsonpath performance

mgwalker:
What exactly is the point of this whole thing, you may be asking. A valid question. It turns out that the performance of `jsonpath` is highly dependent on how generic the search path is. The more generic, the worse the performance. And not by just a little bit. In the case of CARTS data, a query on just an ID (something like `?..(id=='2020-01-a-01-01')`) takes literally **A THOUSAND TIMES** longer than a query for the same node's exact path.

You may have noticed the app seems a little sluggish sometimes. ☝️  That's why. And it got a lot more noticeable with the conditional questions, since they end up doing a lot more jsonpath queries.

So, what if we take all those generic lookups that come in, find the exact path that corresponds to the generic path, and then cache those exact paths? The first queries are still slow, but every subsequent query is suddenly anywhere from 10 to 1000 times faster. So PR creates a drop-in wrapper around `jsonpath` so that we can get the performance benefits everywhere without having to change anything anywhere else, and updates the other bits so that they use this instead of using the `jsonpath` library directly.

### DIFF
--- a/frontend/react/src/components/Utils/helperFunctions.js
+++ b/frontend/react/src/components/Utils/helperFunctions.js
@@ -1,6 +1,5 @@
 // This function extracts just the unique ID
 // from all objectives & goals, stopping at the underscore
-import jsonpath from "jsonpath";
 
 const sliceId = (id) => {
   let idString = id.toString();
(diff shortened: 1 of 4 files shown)


alewoo28 (approved):

alewoo28:
Looks great and this is SIGNIFICANTLY faster! Thanks a lot Greg

PR #556, opened 2020-09-11
Add helper text to subsection

andrewadcock:
Subsection was missing the "text", also known as Helper Text

### DIFF
--- a/frontend/react/src/components/layout/Subsection.js
+++ b/frontend/react/src/components/layout/Subsection.js
@@ -3,10 +3,11 @@ import { connect } from "react-redux";
 import { selectSubsectionTitleAndPartIDs } from "../../store/selectors";
 import Part from "./Part";
 
-const Subsection = ({ partIds, subsectionId, title }) => {
+const Subsection = ({ partIds, subsectionId, title, text }) => {
   return (
     <div id={subsectionId}>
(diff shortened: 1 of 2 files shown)


frontend/react/src/components/layout/Subsection.js
```diff
@@ -3,10 +3,11 @@ import { connect } from "react-redux";
 import { selectSubsectionTitleAndPartIDs } from "../../store/selectors";
 import Part from "./Part";
 
-const Subsection = ({ partIds, subsectionId, title }) => {
+const Subsection = ({ partIds, subsectionId, title, text }) => {
   return (
     <div id={subsectionId}>
       <h2>{title}</h2>
+      {text ? <div className="helper-text">{text}</div> : <></>}
```
mgwalker, line 10:
```suggestion
      {text ? <div className="helper-text">{text}</div> : null}
```

What do you think about the non-text condition not rendering anything at all? I'm not sure what happens if you render an empty fragment, but maybe it's identical to `null`, in which case... ignore. 😄

mgwalker (approved):
🚀

frontend/react/src/components/layout/Subsection.js
andrewadcock: Changed.

PR #560, opened 2020-09-11
Modularize the question component

mgwalker:
- splits the monolithic QuestionComponent into individual components
   - leaves the monolithic QuestionComponent in place so as not to disrupt any ongoing development
   - copies code from there into the individual components rather than reinventing the wheel
   - makes it easier to assign tasks to work on individual components since they're not all embedded in a single larger component
- modifies the conditional display logic: instead of always removing hidden questions, if the question has skip text, replace the question with an ephemeral question object containing the skip text, so we can display it


mgwalker:
Closing as premature. We can revisit in the future.

Issue #522, opened 2020-09-08
FY2020 FMAP Rate 


StephanieCoppel:
As a state user, I want to see the FMAP rates that apply to my state when filling out section 5 (formerly section 4) of the CARTS form.

The FMAP rate varies per state per year.

For this story, we just want to be able to *display* the correct FMAP rate for 2020 in section 5 (formerly section 4)  of the form.  This data will be pre-populated by the dev team, and not editable by users.  Per Tadhg's notes:

-  Add location in Redux Store for FMAP rate. Should be linked to info about the state/state program.
-  Add DB spot for FMAP rates.
-  Add API read endpoint
-  Get FY2020 data. Include it in deployment scripts.

A future release will include DSCP ability to enter these themselves for FY2021 release (#476)

AC: 
- [ ] State users see the correct FY2020 FMAP rate for their state in section 5 (formerly section 4) of the CARTS form

StephanieCoppel:
FMAP rates to be input into the application at link below (scroll down to see the FY2020 rates for each state):
https://www.federalregister.gov/documents/2019/12/03/2019-26207/federal-financial-participation-in-state-assistance-expenditures-federal-matching-shares-for

erinzstrenio:
@StephanieCoppel do you know if the FMAP rates should be pulled from the first or second column from that link?

erinzstrenio:
also just capturing here that DSCP will need to add 2021 FMAP rates onto the 2020 report once they're made available (I believe October?) since states will need to enter their projected 2021 costs onto the 2020 report

StephanieCoppel:
enhanced FMAP rates should be used (the second column) @erinzstrenio

kaitlin:
I submitted a PR for the backend work (bullets #2-4) but the front-end work here to pull it into Redux still needs to happen.

PR #569, opened 2020-09-13
Add 2020 FMAP data to API

kaitlin:
This addresses the backend portion of #522 (https://app.zenhub.com/workspaces/carts-prototype-zenhub-5f0643b2a50cb70019ecc5b0/issues/18f/cms-carts-seds/522).

If you pull this branch down and run the regular docker up/down/rebuild commands, it will load a fixture of the FMAP data. You can query the API for the list:

`/fmap`

or get one state:

`/fmap/GA`

I didn't write any tests (I don't see any backend tests in the project folder but maybe I'm missing them).

For now, I just dump the objects for the FMAP data and reload from the fixture file upon deploy. This would need to change if we get around to allowing admin users to edit the FMAP data but it will do for now I suppose.

Screen Shot 2020-09-12 at 10 22 23 PM


jgillis1 (approved):
This was a lot of work Kaitlin. Thank you for your time on it. The endpoint looks good to me. We do know that we will need to provide a font-end entry mechanism for DSCP in the future.

tadhg-ohiggins (approved):
Looks good to me. Is there a reason plain old `FMAP` isn't required, only `enhanced FMAP`? Just curious, not an issue for the PR.

kaitlin:
@tadhg-ohiggins plain FMAP is used for Medicaid and enhanced FMAP is used for CHIP (Although due to covid and the PHE there is enhanced FMAP in place for Medicaid too I _think_). I loaded both just because it wasn't any harder than adding just one, since they came from the same table.

Issue #470, opened 2020-08-29
Fixtures for section 5 (formerly section 4)


cjhskippy:
Ensure that, upon deployment to a new environment, the content for section 5 is loaded.  Note: this content used to be section 4.

#466 and #518 must be completed first.

A/C:
- [ ] Upon deployment, the questions for section 5 on 2020 reports is populated in the database

Issue #381, opened 2020-08-19
Build of Section 5 


erinzstrenio:
Blocked by issue #518 

As a user, I want to be able to see the questions in section 5 and save my answers.

- [x] Section 5 appears in the navigation bar
- [x] All form content in section 5 displays per [design document](https://docs.google.com/document/d/1N3wgjF4VwMd9HXG7F8xYS13nswctSC03c41HkvGMOus/edit#).
  - [x] Includes column for projected years (get question # this applies to)
- [x] Users can save answers that will be retained when navigating to other parts of the form (e.g. saved in the Redux store)
- [ ] Updates are deployed to the cartsdemo site + pass UAT w/ Stephanie.

Engineering Notes as of 9/25:
- the JSON is probably incorrect for the help text - 
- Completing all of this section may require PR #632 being merged
- if the `show_if` functionality got merged, there’s a question in that section that should be slightly redone in the JSON to make it appear only if other is selected in the checkboxes

Content notes: 

- [x] Add gridview tables for 2020, 2021, 2022 and synthesized tables should have columns for all three years. We consolidated all three years into the questions, so there are now fewer parts and tables. [See content doc](https://docs.google.com/document/d/1pyCyUQPJKVEIKZ3cBTM8ZThxNeLiUzIKPPr0xfImtv4/edit?ts=5f6cea35#) for updates. 

Related to issue #214

kaitlin:
I've started a build out of this on branch `new-section-5`.

jgillis1:
@kaitlin Could you confirm, this ticket is up for grabs to any developer? They can just start with your branch and move forward.

erinzstrenio:
I'm happy to pair with @kaitlin or any other dev who picks this up! The [template doc](https://docs.google.com/document/d/1pyCyUQPJKVEIKZ3cBTM8ZThxNeLiUzIKPPr0xfImtv4/edit?ts=5f6cea35#heading=h.xrab6krx0pon) is the best one to reference for updated language per DSCP edits.

andrewadcock:
Section 5, Part 2, Table 3 relies on FMAP data that currently isn't available. I believe this work has started but may need to be modified so that we can use it with a synthesized table and RPN (i.e. add an ID).

In addition, there is a complex equation which requires multiplying against a sum: FMAP (FFY 2020) x [total program costs]

PR #551, opened 2020-09-10
Move old section 5 to new section 6

mgwalker:
- moves section 5 JSON to section 6
- updates IDs accordingly
- adds section 6 to state JSON
- updates fixtures
- updates (still) temporary JSON the app uses when it can't reach the API
- closes #546

PR #571, opened 2020-09-14
New section 5 JSON and state fixtures

### LINKED ISSUE #466: JSON for section 5 (formerly section 4)
Produce the JSON that will represent section 5.  This used to be section 4 on the old CARTS forms.

A/C:
- [ ] JSON file exists with content for every form element in section 5.

Blocks #470. Part of #381

kaitlin:
Relates to #381 and maybe closes #466 and #470 . 

I created the JSON for (new!) section 5. I also added some fixtures for AZ, AK, and MA. Although, I haven't done anything to the schema data other than change the state value for those fixtures. 

Some issues/areas to look out for:

- There are a few areas where the calculation is more complicated than summing or percentages, and requires a few custom calculations. I made placeholder elements and tried to make notes in the JSON itself, but these are also documented [here](https://docs.google.com/document/d/1N3wgjF4VwMd9HXG7F8xYS13nswctSC03c41HkvGMOus/edit#).
- None of the `synthesized_tables` are showing up despite having the JSON. I don't know if that's me or if this hasn't been implemented on the front end yet. I would need to pair with someone a bit to figure this out.
- There doesn't seem to be a way in the current schema to do a `show_if` instead of a `hide_if`. This meant that for question 23 ("What were your state funding sources in FFY 2020?") where the user can select multiple values, including `other` I don't really have a way to conditionally show the follow up question if _at least_ other is checked. So for now, I just always show the question and say "If you selected `other`, ....".

I also added a few questions for @erinzstrenio in the design doc linked above so some of the text may need to change a bit based on her responses. 

My attempt at a haiku:

section five was four
and section six once was five
but where is four now?


erinzstrenio:
@kaitlin section 4 = formerly section 2B

alewoo28:
Hey Kaitlin, 
I tried looking into maybe adding a helper function for questions like 23, but I dont think theres a way to do it without having the JSON for this question be different from other conditional questions because we're asking it to do the opposite from most conditional questions. 
I think just showing it is fine for now.

mgwalker:
Also replaces PR #551. Will close that one since this also includes the same work.

alewoo28 (approved):
Hey Kaitlin, 
I tried looking into maybe adding a helper function for questions like 23, but I dont think theres a way to do it without having the JSON for this question be different from other conditional questions because we're asking it to do the opposite of most conditional questions. 

This was mentioned a couple months ago as a potential issue and I think @tadhg-ohiggins had a work around for it. 

I think just showing question 10 is a great workaround for now. 

Synthesized tables are part of the question component, they're waiting in Tim's PR for 2b-api (#531 )

mgwalker (approved):
- `synthesized_table` is not yet finished on the frontend; it's not you. I mean, it might be you too, but it's at least not ***just*** you. 😄
- we should definitely add a `show_if`. It wouldn't be too hard. @alewoo28 added a generic `shouldDisplay` utility, and I broke it down by conditional type recently. Plugging in more conditional types ought to be fairly straightforward.

Let's add separate issues to add `show_if` (and edit question 23/4.9) and to also figure out how to handle the case where a value need to be subtracted instead of just added. Those aside, this looks good to go to me, and if we need to update any question text, we can always do it later.

TGriesemer:
@kaitlin  
Just to be clear, synthesized tables are not part of my pr #531.  2b contains Synthesized values and Noninteractive tables. I do not know if someone else is working on synthesized tables but it is not part of 2b.

alewoo28:
My mistake then! Sorry about that.

Issue #542, opened 2020-09-10
Section 3H - numbering issues


StephanieCoppel:
Section 3H begins with questions 1 and 2, restarts at 1 and goes through question 6, and then has questions 1 and 2 again. Numbering should be chronological from 1-10 for this section

andrewadcock:
This appears to be due to incomplete JSON. Once #555 has been addressed this will need to be confirmed.

Issue #527, opened 2020-09-09
Part titles should include Part Number


cannandev:
A/C:
- [x] Part number is displayed in subheadings

Example in 3C: `Eligibility Renewal and Retention` should read `Part 1: Eligibility Renewal and Retention`
Example in [3H](https://docs.google.com/document/d/1_5gBq_96gz3kQ_yQdt09SXI_0Oy8YO5i5YFZZEiGCmk/edit#): `Part 1`

PR #573, opened 2020-09-14
Add Part number prefix to part titles

andrewadcock:
Add `Part n: ` to all part titles.
Generate `n` from partId (already available)

Satisfies AC from #527 and #542

### DIFF
--- a/frontend/react/src/components/layout/Part.js
+++ b/frontend/react/src/components/layout/Part.js
@@ -16,15 +16,12 @@ const showPart = (context_data, programType) => {
 
 const Part = ({ partId, text, title, context_data, programType }) => {
   if (showPart(context_data, programType)) {
-
     // Determine Part Number from partId
     let partNum = Number(partId.split("-").pop());
     return (
       <div id={partId}>
-
         <h2>Part {partNum}{title ? ": " + title : <></>}</h2>
         {text ? <p>{text}</p> : <></>}
-
         <QuestionComponent partId={partId} />
       </div>
     );
--- a/frontend/react/src/components/layout/Part.js
+++ b/frontend/react/src/components/layout/Part.js
@@ -6,9 +6,9 @@ import QuestionComponent from "../fields/QuestionComponent";
 
 const showPart = (context_data, programType) => {
   if (context_data &&
-      programType &&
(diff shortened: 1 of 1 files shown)


jgillis1 (dismissed):

frontend/react/src/components/layout/Part.js
```diff
@@ -6,21 +6,22 @@ import QuestionComponent from "../fields/QuestionComponent";
...
+    !context_data.show_if_state_program_type_in.includes(programType)) {
     return false;
   }
   return true;
 }
 
 const Part = ({ partId, text, title, context_data, programType }) => {
   if (showPart(context_data, programType)) {
+    // Determine Part Number from partId
+    let partNum = Number(partId.split("-").pop());
```
mgwalker, line 20:
```suggestion
    let partNum = partId.split("-").pop();
```

I'd probably skip the number conversion, since it doesn't actually matter. It gets used as a string anyway, so I'd be a little wary about some future where there's a part 3B or something, and then this code throws an exception and we have to hunt it down unexpectedly. 😛

frontend/react/src/components/layout/Part.js
```diff
@@ -6,21 +6,22 @@ import QuestionComponent from "../fields/QuestionComponent";
...
     return (
       <div id={partId}>
-        {title ? <h2>{title}</h2> : <></>}
+        <h2>Part {partNum}{title ? ": " + title : <></>}</h2>
```
mgwalker, line 23:
```suggestion
        <h2>Part {partNum}{title ? `: ${title"` : ''}</h2>
```
Use string interpolation, and make the whole thing a string instead of a mixed string-and-component. (`<></>` is a shortcut for React's built-in `<Fragment>` component, and I'm a little wary of mixing types.)

mgwalker (approved):
Looks good to me. A couple of suggested edits – take them or leave them – but I'm going to mark this as approved and you can merge whenever you're ready!

frontend/react/src/components/layout/Part.js
andrewadcock: I used Number to strip off the leading zero. Is there a better way to handle that?

mgwalker: Ah, that makes sense!

Issue #578, opened 2020-09-14
Checkboxes with conditional logic tied to them


andrewadcock:
Checkboxes are not allowing for dependent children by reference. 

See section 3h, part 2, question 4 (http://localhost:81/sections/2020/03/h?dev=dev-ak) for an example.

This is likely due to the return type of checkboxes being an array as opposed to a string (like radios).

cjhskippy:
Related to (duplicate of?) #567

Issue #555, opened 2020-09-11
Complete JSON for 3h


andrewadcock:
Pulling from the API, the JSON from 3h appears to be missing part titles and helper text.

PR #580, opened 2020-09-14
Update JSON from 3h based on word doc

andrewadcock:
Addresses: #555
JSON has been updated to match output of word doc. Questions, skip text, etc was updated.

A few issues arose during this:
1) Checkboxes with conditional children are not working; child questions are not displaying. I think this is due to return type of the question being an array instead of a string. A new ticket has been added (#578 ).

2) validate_ids.py never finished running, it failed to complete when run over night


jgillis1 (approved):
Worked through the hide_if_2 issue with Andrew and the 2 different conditions on which to show/hide Part 3. This was the simplest solution we came up with and it executes well.
https://docs.google.com/document/d/1_5gBq_96gz3kQ_yQdt09SXI_0Oy8YO5i5YFZZEiGCmk/edit

jgillis1 (approved):
JSON looks good and matches the content doc.
@tadhg-ohiggins @mgwalker This PR is ready for 18F review.

docs/section-schemas/backend-json-section-3.json
```diff
@@ -4098,8 +4096,9 @@
           {
             "id": "2020-03-h-03",
             "type": "part",
+            "text": "Since you didn’t collect the CAHPS survey, go ahead and complete this section.",
             "context_data": {
-              "skip_text": "Part 3 omitted because you indicated that have collected CAHPS survey results",
+              "skip_text": "Part 3 was skipped because it only applies to states that didn’t collect the CAHPS survey.",
```
mgwalker, line 4101: Nothing to fix here, but we should create an issue to update the Part component so that it also hides parts based on `conditional_display` - currently it only hides based on the state program type.

mgwalker (approved):
LGTM!

docs/section-schemas/backend-json-section-3.json
mgwalker: Added #600 for that

PR #583, opened 2020-09-15
WIP: Add conditional logic for parts that rely on questions

andrewadcock:
Previously, parts were only conditional based on program type. 3h requires parts that are dependent on question answers. This satisfies that need.


andrewadcock:
Taken care of in another PR.

PR #586, opened 2020-09-15
Delete some unused code

mgwalker:
- a lot of the static section content has been replaced with dynamic sections and questions, so we can clean out the lingering static sections
- we can also remove those static sections from the router, and hook up the static Section 2B to use the same route structure as the fully dynamic routes
- before the API was ready, the various sections loaded data from local JSON; since they are now getting data from the redux store, we don't need that local JSON, so we can delete that too!


frontend/react/src/components/layout/TOC.js
```diff
@@ -1,160 +1,94 @@
 import React, { Component } from "react";
```
mgwalker, line 1: I didn't actually change anything in this file. I deleted `TOC.js` and renamed `TOC-dynamic.js`, but I guess since I did it all in one commit, git doesn't know the difference.

jgillis1 (approved):
Lots of moving parts here but it looks like good clean up to me.

mgwalker:
Gonna hold this until #531 so it doesn't create a lot of merge conflicts on that one. I think they'll be easier to clean up on this side.

PR #593, opened 2020-09-15
Minor README update

tadhg-ohiggins:
Don't forget data_net
We're lost without a network
So we create one.

Add the step `docker create network data_net` to setup instructions. Strike bit about `local-additional.sh` as that should be covered by the docker-compose entrypoint now.


mgwalker (approved):

mgwalker:
"Don't forget data_net" is 6 syllables, but I'll let it slide this time.

Issue #553, opened 2020-09-10
JSON and Fixtures for Section 3i


cjhskippy:
Subtask of #382 . Ensure that the content for section 3i is represented in JSON.

- [x] Build JSON for section 3i
- [x] Add fixtures for the new section so the JSON is deployed automatically

@jgillis1 - @TGriesemer can take over build portion.

PR #602, opened 2020-09-16
Feature/3i

jgillis1:
JSON and Fixtures for Subsection 3i. The front end renders everything but the repeatables, which still needs to be built out and the synthesized value (after P2-Q5), which is pending Section 2b's code. They will be addressed in the 3i build ticket.

Content Doc: https://docs.google.com/document/d/1dRgcKDL7tyGJ3EQuBomhW03xxyeQQ-Hv3eqSNnouutw/edit

Issue #553


andrewadcock (commented):
Looks good. Once the repeatables are built out, this should work fine!

cl-adavyd (approved):

mblevins-collabralink (approved):

jgillis1:

docs/section-schemas/backend-section.schema.json
```diff
@@ -64,6 +64,9 @@
         "text": {
           "type": "string"
         },
+        "text_2": {
+          "type": "string"
+        },
```
mgwalker, line 69: I'm not comfortable with `text_2` to handle a second paragraph. It has the potential to get out of hand and doesn't scale at all. Two alternatives that come immediately to mind:

1. Add newline characters to the text and use those as paragraph delimiters. When the text is rendered, split it on newlines and then map the pieces to paragraphs.
   ```js
   text.split("\n").map(paragraph => <p>{paragraph}</p>)
   ```
   The upside is that we wouldn't have to change any of our components that use the `text` property – they'd render the newline literally and it'd be a harmless single whitespace. We also wouldn't have to update any other subsection data since it's a convention rather than a schema change.

2. Turn the `text` property into an array. Map the array elements into paragraphs.
   ```js
   text.map(paragraph => <p>{paragraph}</p>
   ``
   The upside to this approach is I think it's more explicit in the data how the text is meant to be rendered. The downside is we'd have to update every existing subsection in the JSON since the schema changed. Alternatively, `text` could be updated so that is *one of* an array of strings or a single string, which just creates a little bit of extra complexity in parsing it.

@tadhg-ohiggins Do you have thoughts or preferences about the best way to handle paragraphs in subsection helper text?

frontend/react/src/components/layout/Part.js
```diff
@@ -49,7 +54,8 @@ const mapStateToProps = (state, { partId }) => {
     title: part ? part.title : null,
     context_data: _.has(part, "context_data") ? part.context_data : null,
     programType: state.stateUser.programType,
+    state: state,
```
mgwalker, line 57: I addressed this in [#604](https://github.com/18F/cms-carts-seds/pull/604#discussion_r490566765), but I won't rehash it here. I assume it'll get worked out in a rebase/merge later.

mgwalker (changes requested):
Let's talk about the schema change in more detail. Also, please update the schema documentation if you make changes to the schema. (It's the ReStructured Text `backend-json-structure-documentation.rst` file in the `docs/section-schemas` directory.)

docs/section-schemas/backend-section.schema.json
jgillis1: Gotcha. So, to clarify, when an existing property doesn't meet the content needs, it's acceptable to modify the property accordingly rather than create a new property?
I think I prefer option 1, personally so I will look into that tomorrow while Tadhg considers them as well.

frontend/react/src/components/layout/Part.js
jgillis1: Yes, however it's resolved in #604 will take precedent. I just needed something in this branch to be sure the JSON was complete.

docs/section-schemas/backend-section.schema.json
tadhg-ohiggins: I prefer the first option since the cases where this is necessary seem to be the minority. If we find later that most question text has newlines, we should switch to an array as a future enhancement.

tadhg-ohiggins: Also, either changing or adding properties probably means the issue is worth bringing up for discussion.

mblevins-collabralink: I concur that option one is better at this point - and I should have caught this when reviewing it.

mgwalker (approved):
A minor typo and a question about structuring the data, but nothing I see that's worth holding up the merge!

A question for @erinzstrenio: in the docs, the language is to "skip to the next section" if they answer no to part 1, question 1. I think the implementation here follows the docs, but the upshot is that parts 2 and 3 will both show a skip-text alert if the answer to question 1 is no. Should we create an issue to make it such that there is only one skip-text alert shown? I don't *thing* there's a way in the current JSON schema to hide multiple parts with a single skip-text, but we can investigate.

erinzstrenio:
@mgwalker We can actually get rid of Part 3 and move that single question from Part 3 to the end of Part 2 to avoid the double skip-text alerts. Does that help? (hopefully didn't make it more complicated?) Just updated the google doc.

frontend/react/src/components/layout/Subsection.js
```diff
@@ -3,11 +3,11 @@ import { connect } from "react-redux";
 import { selectSubsectionTitleAndPartIDs } from "../../store/selectors";
 import Part from "./Part";
 
-const Subsection = ({ partIds, subsectionId, title, text }) => {
+const Subsection = ({ partIds, subsectionId, title, text, text_2 }) => {
```
mgwalker, line 6:
```suggestion
const Subsection = ({ partIds, subsectionId, title, text }) => {
```

frontend/react/src/components/layout/Subsection.js
```diff
@@ -24,6 +24,7 @@ const mapStateToProps = (state, ownProps) => {
     partIds: subsection ? subsection.parts : [],
     title: subsection ? subsection.title : null,
     text: subsection ? subsection.text : null,
+    text_2: subsection ? subsection.text_2 : null,
```
mgwalker, line 27:
```suggestion
```

frontend/react/src/store/selectors.js
```diff
@@ -21,6 +21,7 @@ export const selectSubsectionTitleAndPartIDs = (state, subsectionId) => {
       parts: subsection.parts.map((part) => part.id),
       title: subsection.title,
       text: subsection.text,
+      text_2: subsection.text_2,
```
mgwalker, line 24:
```suggestion
```

mgwalker (commented):
Oops! Forgot to mention these earlier, but for some reason standup brought it back to mind.

docs/section-schemas/2020-ak-section-3.json
```diff
@@ -4182,6 +4182,318 @@
             ]
           }
         ]
+      },
+      {
+        "id": "2020-03-i",
+        "title": "Health Services Initialtive (HSI) Programs",
```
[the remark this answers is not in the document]
jgillis1: Typo has been corrected and the content changes Erin put up on 9/18/2020 have been made.

Issue #600, opened 2020-09-16
Update Part component to allow conditional hiding


mgwalker:
Currently the Part component can only be hidden based on the state's program type. Parts can also be hidden conditionally based on the answers to questions. We need to update the Part component accordingly.

Issue #584, opened 2020-09-15
Skipped Parts message should be more prominent


andrewadcock:
Currently, skipped parts render a <p> with the skip text. This should be more prominent, similar to <Alert />

PR #604, opened 2020-09-16
Check for conditional when rendering parts

andrewadcock:
Previously, parts were only hidden based on programType. This allows parts to check shouldDisplay for parts.

Ticket #600, #584

### DIFF
--- a/frontend/react/src/components/layout/Part.js
+++ b/frontend/react/src/components/layout/Part.js
@@ -21,11 +21,11 @@ const showPart = (context_data, programType, state) => {
   return true;
 }
 
-const Part = ({ partId, text, title, context_data, programType, state }) => {
+const Part = ({ partId, text, title, context_data, show }) => {
   // Determine Part Number from partId
   let partNum = Number(partId.split("-").pop());
 
-  if (showPart(context_data, programType, state)) {
+  if (show) {
     return (
       <div id={partId}>
         <h2>Part {partNum}{title ? ": " + title : null}</h2>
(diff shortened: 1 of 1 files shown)


frontend/react/src/components/layout/Part.js
```diff
@@ -49,6 +55,7 @@ const mapStateToProps = (state, { partId }) => {
     title: part ? part.title : null,
     context_data: _.has(part, "context_data") ? part.context_data : null,
     programType: state.stateUser.programType,
+    state: state,
```
jgillis1, line 58: Is it appropriate to return the full state?

jgillis1 (changes requested):

frontend/react/src/components/layout/Part.js
andrewadcock: It's a requirement for shouldDisplay.

jgillis1 (approved):

cl-adavyd (approved):

jgillis1:

frontend/react/src/components/layout/Part.js
mgwalker: It is not appropriate to return the full state for performance reasons, unfortunately. This will mean that any change anywhere in the state will cause every Part to re-render, which can be expensive.

Instead, you could move the `showPart` check into the `mapStateToProps` and just return a boolean instead of the whole state. E.g.:

```js
const mapStateToProps = (state, { partId }) => {
  const part = selectFragment(state, partId);
  const contextData =  _.has(part, "context_data") ? part.context_data : null;

  return {
    text: part ? part.text : null,
    title: part ? part.title : null,
    context_data: contextData,
    programType: state.stateUser.programType,
    show: showPart(contextData, state.stateUser.programType, state)
};
```

Then in the component, you'd extract `show` from the props and use that as the conditional.

mgwalker (changes requested):

jgillis1 (approved):

mgwalker (approved):
Nice

Issue #475, opened 2020-08-31
Form saves to database


cjhskippy:
As a state user, I want my responses to be saved as I progress through the form so that I don't have to enter the entire report in one marathon session.

As a state user, I only want authorized users to be able to save changes to my state's report.

Current design choice for this interaction may need discussion: will we auto save or have a save button?  (see discussion in #256 )

A/C:
- [x] Only authorized users are able to save form responses.
- [x] Users are notified if there is an error saving the form
- [x] Users are provided visible updates when the form is saved.
- Replaced with #783 ~Answers provided by user are saved to persistent storage.  Upon logging out and visiting the same page a second time, the saved answers are loaded.~

Depends on user authentication #392

mgwalker:
The remaining acceptance criteria are blocked by getting auth tied in on the backend side. The only frontend change that should be necessary will be having it post data to the real URL (instead of the dummy one created for this PR).

timoballard:
#705 is ready for review/merge. Adds saving functionality to the API without authentication or authorization checks, as those are being implemented more broadly as separate items (see #736 and #737)

timoballard:
This was mistakenly autoclosed by #705 - reopening

jgillis1:
@timoballard Could you confirm the saving is occurring? I was on mdctcartsdev.cms.gov and attempted to change data, navigated to another site, came back and my change wasn't shown. After clicking around on the page a few times, this error message popped up:
![image.png](https://images.zenhubusercontent.com/5f11af98ebec4c4565352e89/da251b9f-e81e-49cd-8a4f-b3e69af2350f)

mgwalker:
We're removing the acceptance criteria related to the form status, right? In which case, since saving is working, it looks like all that's left is the "authorized users" piece. Is that right @timoballard?

@StephanieCoppel If ☝️  is right, I think moving this to MGR review would be appropriate, and we can have a separate issue for making sure that only authorized users can actually do the save. I suspect a whole lot of auth stuff will turn on all at once. It's a work tracking question, so however you want to proceed – but the form is saving to the database now. 😄

timoballard:
That’s correct. The auth portion is progressing more generally in #736. The plan is to integrate that with @tadhg-ohiggins user permissions work (#737), at which point a lot of this should snap into place

Issue #256, opened 2020-07-31
Interaction tip from eAPD for auto-save


austinhernandez:
All of the users from the most recent round of usability testing had difficulty finding and understanding the save interaction. The disabled text link color contrast is too low, and there is a mistrust from previous versions that work is saved. 

I suggest using the same code from eAPD's save button. During their usability testing, they found that users noticed the autosave functionality when there was an animation duration of 700ms between save and auto save buttons.

cjhskippy:
Was no longer needed given the save behavior implemented so far.

PR #612, opened 2020-09-17
Save UI

mgwalker:
- addresses half of #475
- adds a "Saving" status to the header, along with the date and relative time of the last save
- catches save errors and displays a warning to the user
- sends the correct data to the API for saves
- adds a dummy endpoint to the API for saving

When the user has entered or changed some data and the autosave process is triggered, the header is now updated with an animated spinner, which will spin until the save has finished. (Note that the spinner will always be visible for at least 750 ms, even if the save finishes faster. eAPD user research showed that users would often not even notice shorter animations and were unaware that their document had been saved. Mentioned in #256 too.)

### Successful save

When the save is complete, the header switches back to showing the most recent successful save time:

success

- For less than a minute, no time is shown
- For less than a day, the last save time and relative time will be shown:
   image
- For less than a year, the last save ***date*** and relative time will be shown (e.g., `July 17 (2 months ago)`)
- For more than a year, the last save date is used (e.g., `July 17, 2019 (1 year ago)`)

Hopefully I didn't overstep by putting this behavior in, but this comes from the user-tested autosave UI from eAPD so I thought it made a good starting point, at least.

### Unsuccessful save

If the save fails, this error message pops up:

error

### Known shortcomings

I think these are just FYI for @StephanieCoppel for future things for possible future refinements.

- The header in eAPD is stuck to the top of the screen so the "saving"/"last saved" information is always visible. The header in CARTS scrolls off the top of the page, so a user who is further down the page may not have any indication that their document has been saved without scrolling back up to look at it.
   - As an aside, making the header stick to the top of the screen is a bigger lift than it seems like it should be. It's especially troublesome if CARTS needs to fully support Internet Explorer. eAPD made the decision to "partially" support IE, which allowed us to ignore its shortcomings and let it just be ugly.
- The error dialog from eAPD was designed to stick to the bottom of the header, which is always visible. For CARTS, the header scrolls off the top of the page, and the error dialog can become detached and float weirdly over the user's work.

### DIFF
--- a/frontend/api_postgres/carts/carts_api/views.py
+++ b/frontend/api_postgres/carts/carts_api/views.py
@@ -112,6 +112,9 @@ def sections_by_year_and_state(request, year, state):
                                        context={"request": request})
         return Response(serializer.data)
 
+@api_view(["POST"])
+def temp_post_endpoint(request, year, state):
+    return HttpResponse(status=204)
 
 @api_view(["GET"])
 def section_by_year_and_state(request, year, state, section):
--- a/frontend/api_postgres/carts/urls.py
+++ b/frontend/api_postgres/carts/urls.py
@@ -29,6 +29,8 @@ router.register(r'fmap', views.FMAPViewSet)
 api_patterns = [
     path("sections/<int:year>/<str:state>",
          views.sections_by_year_and_state),
+    path("sections/<int:year>/<str:state>/temp",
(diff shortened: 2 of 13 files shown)


tadhg-ohiggins (approved):

StephanieCoppel:
@mgwalker appreciate the heads up on functionality/future refinements!!!

PR #616, opened 2020-09-18
ENABLE-219: Prod values for DMS job

sethsacher:


tools/dms/prod/aws/dms_task.tf
```diff
@@ -11,7 +11,7 @@ module "dms_prod" {
   source_database_name_seds  = "SCHIP"
   source_database_username = "mbescbes"
   source_database_password = "${var.source_database_password}"
-  source_database_host     = "source_database_host"
+  source_database_host     = "192.168.174.25"
```
dwhitecl, line 14: I'm thinking this should be a variable set similar to PW as thats a real IP

dwhitestratiform (approved):

Issue #477, opened 2020-08-31
American Community Survey data support


tadhg-ohiggins:
State users need to be able to see the American Community Survey data in the start of Section 2.
The data varies per state per year.

For this issue, the data does not need to be user-editable.  That use case is covered in #637 

AC:
- [x] Add location in Redux Store for this. Should be linked to info about the state/state program.
- [x] Add DB spot for this.
- [x] Add API read endpoint.
- [x] Add API write endpoint.
- [x] Add permissions for endpoint.
- [x] Add frontend UI for endpoint OR get data from DSCP directly so that we can enter it, if the frontend for central office users isn't done by then.
- [x] Table in Section 2 Part 2 (first table) is populated with ACS data for the last 5 years (based on example in content doc)
- [x] JSON is updated appropriately to display the data from the new endpoint
- [x] Update README with how this data is loaded to Section 2 Part 2

kaitlin:
This is probably something I can knock out (at least all the backend bits) if it's needed for release. I noticed in 2020-02-a-02 in the base JSON template that it's prefilled with dummy data. We may want to update that JSON to be empty by default in the course of this ticket. 

I also don't think we need permissions on this endpoint, since it's public data anyway?

kaitlin:
Just as a note, I'm noticing that the [HI-10 tables](https://www.census.gov/data/tables/time-series/demo/health-insurance/acs-hi.2019.html) that DSCP uses don't include data for American Samoa (AS), Guam (GU), or the Northern Mariana Islands (MP). They do have data for DC and Puerto Rico. It may not be an issue since I know at least in AS, everyone is eligible for Medicaid so the uninsured rate might not be meaningful.

heymatthenry:
These data made it into Redux along with the FMAP data as part of #717. The data aren't surfaced in the JSON, but each state's ACS data are there in Redux under `allStatesData`.

PR #617, opened 2020-09-21
Populating ACS data for uninsured children in API - moving FMAP

kaitlin:
Resolves the bottom 5 checkboxes in the list in #477 . This PR does a few things:

- Creates a stubbed out state model in the database and API (just name and code/abbrev for now but could store other things here)
- Creates an ACS data model, loads ACS data on uninsured children under 200% FPL for 2014-2019 from fixtures (original, cleaned CSVs also included)
- Removes the previous FMAP endpoint, and instead loads that data (and the ACS) data as part of the new state endpoint (e.g. `http://localhost:8000/state/AK/`)
- Removes what seems like an extra `generate_fixtures.py` file in the management command directory (@tadhg-ohiggins let me know if this is being used somewhere and I can restore)
- Tidies up some URL routing, groups the data loading in `docs/generate_fixtures.py` into functions
- Adds a minimal test file for the state API endpoint

Screen Shot 2020-09-20 at 10 55 25 PM


cl-adavyd (commented):
Good afternoon,

is there a reason why numerical data is represented via different data types?
For example, number_uninsured is an integer, while percent_uninsured is a string.

kaitlin:
@tonydavydets because integers are represented the same everywhere, but I didn't want a fixed-precision decimal value to be converted to a floating point number inadvertently. AFAIK it's not actually used in a calculation -- just a reference table for the user.

tonydavydets:
> @tonydavydets because integers are represented the same everywhere, but I didn't want a fixed-precision decimal value to be converted to a floating point number inadvertently. AFAIK it's not actually used in a calculation -- just a reference table for the user.

tonydavydets:
I see.  Could we please change all numerical data to be represented as numbers / floats?

Having a true representation of data (meaning, integers are represented as integers, floats as floats, etc.) gives us more uniformity, takes guessing out of the question (is a number truly a number or a string), and lastly, gives us the ability to reformat display with one line of configurable code as opposed to changing the entirety of our data / doing extensive regression testing, etc.

kaitlin:
Hmm I see what you're saying, and it does seem like it would make it more consistent with how this data is represented in the dummy state section data. In this case, I'm relying on the default serialization behavior of Django Rest Framework, which has attempted to set a sane default based on the majority of use cases (and presumably because of the float vs Decimal precision issue). For the numbers we're dealing with I think it's fine, but the only way to change this without digging into the internals of DRF is to [use a setting in `settings.py`](https://www.django-rest-framework.org/api-guide/settings/#coerce_decimal_to_string) that would enforce this behavior at a project level. That probably won't be an issue since I think so far I'm the only person who has serialized from any Python objects that aren't primitive types. I'll push a commit now.  Thanks!

cl-adavyd (approved):

PR #619, opened 2020-09-21
DSCP content edits to sections 1, 2, and 3

erinzstrenio:


docs/section-schemas/backend-json-section-0.json
```diff
@@ -6,7 +6,7 @@
     "ordinal": 0,
     "id": "2020-00",
     "type": "section",
-    "title": "Carts Basic Info",
+    "title": "Basic State Information",
```
tadhg-ohiggins, line 9: In the state-specific sections, this is “Info”, but here it's “Information”. (Yet another reason to just generate the example state-specific sections from the generic sections.)

tadhg-ohiggins (commented):
Looks fine to me apart from that one comment.

tadhg-ohiggins:
The files created after `docs/section-schemas/generate_fixtures.py` is run should probably also be part of this PR; we can add them once the other edits are finished.

docs/section-schemas/backend-json-section-0.json
erinzstrenio: Yes - I updated it to "Information" (and yes, agreed on generating the updates for state sections so we don't have to manually update every time!)

tadhg-ohiggins (approved):
Merging with the caveat that there were merge conflicts in the fixture files but not the files in `/docs/section-schemas`. There are a few possible reasons for this and so it may be nothing, but we should be watchful for reversions in Section 3.

PR #607, opened 2020-09-17
Feature/inputgrid component

cannandev:

PR #621, opened 2020-09-21
Modularize the question component

mgwalker:
When I was working on the input grid component, I was having trouble slotting into the monolithic question component, so I took some time to split it up into modules. This preserves the logic and behavior that was previously implemented in `QuestionComponent`, `CMSChoice`, `CMSRanges`, and `CMSRange`, but restructures it so that it is more dynamic and flexible.

I also added [eslint](https://npm.im/eslint) with Airbnb+prettier rules. It's not enabled globally because that's a big, heavy lift, but I did apply it to the code I was restructuring. As a result, I tweaked some of the code to satisfy the linter.

### Added behavior

- Objectives, goals, and other repeatables work (including Section 2B; pulled mostly from the logic in #531)
- Synthesized values work
- Synthesized tables work (pulled from #607)
- Updates the router to be entirely dynamic
- Skip text for hidden questions that contain it (parts were already handled)

### Known issues

- Section 5 needs to be able to synthesize a value in a more complex way than we currently handle. As a stopgap, the Section 5 JSON includes an invalid target ID, causing it to fail to render. This should be pretty minor to fix.


frontend/react/.env
```diff
@@ -1,2 +1,3 @@
+SKIP_PREFLIGHT_CHECK=true
```
mgwalker, line 1: create-react-app comes with its own version of eslint. If you install a newer version of your own, the app will refuse to build (in dev only, presumably). It's a little silly. Anyway, this disables the preflight version check so that it'll build in dev and we can keep going.

frontend/react/src/components/fields/Autosave.js
```diff
@@ -1,23 +1,20 @@
-import React, { Component } from "react";
+import React from "react";
```
mgwalker, line 1: For a few of these, I just tweaked them a bit to satisfy the linter. I was running it on this whole `fields` directory.

frontend/react/src/components/fields/DateRange.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
```
mgwalker, line 1: This component is just a thin wrapper over the existing `DateRange` component, to basically glue up the two `onChange` function styles.

jgillis1:
@mgwalker Would you have time tomorrow or Thursday to meet with the CL team on your changes? We would like to develop a more complete understanding of what has been implemented, why and what, if anything is still outstanding? If a meeting isn't convenient, perhaps you could provide a write up? We want to ensure we are 100% on the same page so we can continue to provide the best possible product to CMS.

mgwalker:
Yes, I can definitely make time for that! I'll message you on Slack to see if we can find a time, and if there's not a good time, I'll write something up to go into more detail.

frontend/react/src/components/fields/Radio.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  return question.answer.options.map(({ label, value }) => (
+    <Choice
+      checked={checked === value}
+      type="radio"
+      value={value}
+      {...childProps}
+      {...props}
+      onChange={onCheck}
```
jgillis1, line 34: We need a disabled property here `disabled={question.answer.readonly}`
Examples: Basic Info Part 1 Q2

frontend/react/src/actions/repeatables.js
```diff
@@ -0,0 +1,70 @@
...
+  delete newItem.questions[0].answer.default_entry;
+  delete newItem.questions[0].answer.readonly;
+  // Truncate the goals down to just one
+  newItem.questions[1].questions.length = 1;
+
+  dispatch({
+    type: SET_FRAGMENT,
+    id: parentId,
+    value: { ...parent, questions: [...parent.questions, newItem] },
+  });
```
jgillis1, line 69: Again, I'm just really impressed with the simplicity you were able to achieve here.

frontend/react/src/components/fields/Percentage.js
```diff
@@ -0,0 +1,50 @@
...
+        className="ds-c-input"
+        errorMessage={error}
+        label=""
+        name={question.id}
+        numeric
+        onChange={change}
+        value={question.answer.entry || ""}
+        {...props}
+      />
+      %
```
jgillis1, line 40: The % is displaying on the line below the response. Can we get it on the same line or perhaps in the text box?

frontend/react/src/components/fields/SynthesizedValue.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
```
jgillis1, line 1: This (I believe it's this) is appearing on screen as a disabled text box. Can it not be a text box and just show/not show the value?
Examples: 3C Part 3 Q4 and 3I Part 2 Q5

frontend/react/src/components/fields/Repeatables.js
```diff
@@ -0,0 +1,62 @@
+import React from "react";
```
jgillis1, line 1: General question and it may not go here but, when a Repeatable is added, it's not immediately apparent on the screen that anything happened (other than the side scroll bar gets smaller). Is it possible to have the screen focus on the first element of the new Repeatable?

frontend/react/src/components/fields/Objectives.js
```diff
@@ -0,0 +1,58 @@
+import React from "react";
```
jgillis1, line 1: Same question here as with repeatables. General question and it may not go here but, when an Objective is added, it's not immediately apparent on the screen that anything happened (other than the side scroll bar gets smaller). Is it possible to have the screen focus on the first element of the new Objective?

jgillis1 (changes requested):
- [x] `hint={question.hint}` is missing from it looks like all fields/component.js files `{...props}` is not executing it (if that is supposed to do so)

- 3C Table: Denials for First Time Applicants looks good and is calculating nicely

frontend/react/src/components/fields/Percentage.js
mgwalker: I want to spin this into a separate issue. The CMS eAPD app has a percent input type that puts the % inside the textbox, like with the currency text fields. I haven't looked to see how much effort it'd be to port that over to CARTS. Let me take a look, and if it's more than a quick fix, would you be okay with merging this and fixing the percent display separately?

frontend/react/src/components/fields/Objectives.js
mgwalker: Hard agree on both of these. We also need to do that for screenreader users anyway.

frontend/react/src/components/fields/Percentage.js
mgwalker: Never mind, it wasn't that much effort. 😄

cl-adavyd (commented):
a slightly cleaner way of making sure an integer less than 10 is preceded by a 0 is to use the javascript padStart function.

Example:

lastPart < 10 ? "0": "" is the same as
lastPart.toString().padStart(2, '0')

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

so, entire incrementId function can be re-written as one line of code:

const incrementId = (id) => {
   return parseInt(id.split("-").pop()) + 1).toString().padStart(2, "0"); 
}

frontend/react/src/components/fields/Integer.js
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from "react";
+import PropTypes from "prop-types";
+import { TextField } from "@cmsgov/design-system-core";
+
+const Integer = ({ onChange, question, ...props }) => {
+  const [error, setError] = useState(false);
+
+  const change = ({ target: { name, value } }) => {
+    const numeric = +value;
+    const parsed = parseInt(numeric, 10);
```
tonydavydets, line 10: lines 9 and 10 seem redundant.  either one will do

frontend/react/src/components/fields/Question.js
```diff
@@ -0,0 +1,123 @@
...
+  ["phone_number", PhoneNumber],
+  ["radio", Radio],
+  ["ranges", Ranges],
+  ["repeatables", Repeatables],
+  ["skip_text", SkipText],
+  ["text", Text],
+  ["text_medium", TextMedium],
+  ["text_multiline", TextMultiline],
+  ["text_small", TextSmall],
+]);
```
tonydavydets, line 49: I may be wrong, but I think I see a pattern... couldn't we infer the value of the map by:

* splitting the key string by '_'
* init capping each array member
* joining together by "" (blank) character

this would eliminate the need to hardcode map like that.  

what do you guys think?

mgwalker: In a scenario like this, I prefer being explicit over inferring. The inference incorporates assumptions and logic that future developers won't need to bother with if we're just explicit.

frontend/react/src/components/fields/Integer.js
mgwalker: They're not redundant. The `+` unary operator will convert to a number – including floating point – but `parseInt` will convert to an integer. This difference is necessary to catch the situation where the user types `1.`. If we only use one or the other, that will pass and we won't display an error. However, the UI won't update to show the `.`. Since we're eating user input, we need to show an error, so we need to handle that condition.

frontend/react/src/components/layout/PageInfo.js
```diff
@@ -1,18 +1,16 @@
...
-      </div>
-    );
-  }
-}
+const PageInfo = ({ name }) => (
+  <div className="page-info">
+    <div className="edit-info">Draft | Last Edit: 4/3/20</div>
+    <h1>
+      {name} CARTS{} FY2020
+    </h1>
```
tonydavydets, line 10: Should we make both dates configurable (stored in props / state) as opposed to hardcoding in template source?

mgwalker:
@tonydavydets 

```
parseInt(id.split("-").pop()) + 1).toString().padStart(2, "0");
```

The `pop()`ed value is a string, so that `+ 1` will be the same string with the number 1 appended to it. The popped value needs to be converted to a number before doing math on it.

I like the suggestion to use `padStart`!

frontend/react/src/components/layout/PageInfo.js
mgwalker: Eventually, definitely. But out of scope of this PR.

cl-adavyd (commented):
other than a few minor things / questions I pointed out, looks great overall.  I don't think any of them will drastically affect functionality; more something for us to think about from the point of standardizing / optimizing.  looks like a ton of work went into it.  seems very well thought out.

tonydavydets:
@mgwalker 
"...Conciseness is not always more readable or obvious, and I prefer the latter..."

couldn't agree more lol.  wasn't sure what we were going for here (sometimes less lines is called for).

but I agree 100%

frontend/react/src/components/fields/Question.js
tonydavydets: ok, sounds good.  so, what I'm getting overall is, we're shooting for clarity over reducing lines of code.

will keep it in mind when writing / reviewing code 👍

cl-adavyd (approved):

jgillis1 (approved):

mgwalker:
Making a note here: I was going to make the switch over to `String.padStart`, but it's not supported in IE. I don't know what our browser support requirements are or what polyfills we're using, so for now, I'm just going to leave it alone.

alewoo28:
Hey Greg, 
I pulled this branch down and tested functionality.I did my best to look through all 53 files but its a lot of code. 

I want to double check the functionality of the DateRange. 
I'm not able to edit any of the fields after all 4 have been entered. Have you seen this issue as well? (Section 2a, question 3B)

I'm also having some trouble with the percentage fields. (Section 3C, question 1a). I can only enter and delete one character at a time before having to click away and click back to be able to type.

mgwalker:
I'll take a look at those two issues first thing tomorrow!

mgwalker:
I've confirmed the DateRange issue. Looks like it's in the old DateRange component (or something going wonky in the combination of the old and new ones).

PR #629, opened 2020-09-22
ability to override okta authentication for localhost / cartsdemo

cl-adavyd:


jgillis1 (approved):

mblevins-collabralink:
A reasonable hack.

mblevins-collabralink (approved):
Should be safe on localhost.

jgillis1:

frontend/react/src/wrapSecurity.js
```diff
@@ -10,21 +10,33 @@ import InitialDataLoad from "./components/Utils/InitialDataLoad";
...
+    window.location.pathname.split("/")[1] === "coming-soon" ? null : <Footer />;
   const loc = qs.parse(useLocation().search);
   const devKeys = { "dev-ak": "AK", "dev-az": "AZ", "dev-ma": "MA" }
-  if (loc.dev && Object.keys(devKeys).includes(loc.dev)) {
-    const userData = {userToken: loc.dev}
-
+  // *** !!!! THIS PART WAS ADDED / SLIGHTLY AUGMENTED
+  // *** pull apart the url using '/' as a separating delimiter
+  const urlParts = window.location.href.trim().toLowerCase().split('/');
+  const domain = urlParts[2];
```
mgwalker, line 25:
```suggestion
  // *** !!!! THIS PART WAS ADDED / SLIGHTLY AUGMENTED
  // *** pull apart the url using '/' as a separating delimiter
  const domain = window.location.host;
```

`window.location` is a [Location object](https://developer.mozilla.org/en-US/docs/Web/API/Location/host) and it already conveniently has properties for the various bits of the location. 🙂

Though as I think about it more, I'm wondering if instead of checking `localhost` for the domain, maybe it'd be better to check if `process.env.NODE_ENV === 'development'`. It's provided automatically by create-react-app and it'll also cover the case where someone happens to navigate locally using `127.0.0.1` or whatever else.

You'd still need to check the domain for cartsdemo, but I'm not sure we should bypass authentication on cartsdemo. Since it will be writing to a live database, I'm concerned about a bad actor finding the site, putting in profane or offensive content, and then someone else being exposed to it.

mgwalker (changes requested):
Aside from parsing out the domain, the implementation here is fine by me.

Open question about whether we should bypass auth altogether on cartsdemo, which I think is one we should get @StephanieCoppel or @kaitlin to weigh in on.

Marking request changes for now and we can resolve it once we've got a clear answer!

frontend/react/src/wrapSecurity.js
tonydavydets: yeah sure.  this was just a quick hack I came up with (more for devs really).  thoughts anyone?

PR #632, opened 2020-09-22
Support more complex synthesized values

mgwalker:
- adds an `rpn` action type, that allows defining a [Reverse Polish Notation](https://en.wikipedia.org/wiki/Reverse_Polish_notation) expression
- updates the `synthesizeValue` helper utility to parse RPN expressions
- updates the JSON and fixtures for section 5 to use RPN for the somewhat more complex synthesized values


frontend/react/src/util/synthesize.js
```diff
@@ -33,10 +36,62 @@ const percent = ([numerator, denominator], precision = 2) => {
   return null;
 };
 
+const rpn = (targets, rpnString) => {
```
tadhg-ohiggins, line 39: I like it! Can we change `targets` here to `values` to make clearer that the JSON lookup of the literal entries for `targets` has to happen outside this function?
I wonder if this function is one where we do need tests, too, in part because I don't think RPN itself is intuitive for people who aren't familiar with it.

heymatthenry: Not a dealbreaker by any means but I'm inclined to agree with the suggestion to change `targets` to `values`. I'm also on board with the suggestion to add some tests for this: 1) For the reason Tadhg mentioned; but also because 2) to the extent that the parser isn't _truly_ doing RPN, such as allowing numeric operands among/after the list of operators, a test case demonstrating that behavior might be helpful.

kaitlin:
I noticed that the synthesized tables don't format the same way as the input fields (i.e. using thousands separators). It would be nice if it did that so it was more readable. Should it carry the `type` attribute forward from the source value?

Screen Shot 2020-09-24 at 3 51 57 PM

kaitlin:
also, it would be good to format the calculated percentage to just 1 or 2 decimal places

Screen Shot 2020-09-24 at 4 03 26 PM

mgwalker:
Added some tests, which revealed some bugs (hooray for tests!).

@kaitlin I like that idea. I think it'll require a change in the JSON to give each computed value a type, so I'd suggest we create a new issue for that. It shouldn't be a huge effort, but it's... not none, either.

heymatthenry (approved):

PR #633, opened 2020-09-23
Okta end to end with fake users

### LINKED ISSUE #481: Authentication: React can accept user information from Django
In order to make the frontend aware of user's information (like the state they represent) or their level of permission, React needs to be able to query Django for authentication details.

- [x] Django API can accept a token from React and verify its validity
- [x] API returns information corresponding to a user (name, state, user role) when a valid token is provided
- [x] API rejects invalid tokens / missing tokens
- [ ] React app queries API on login and can print out results (even if printed to console)

tadhg-ohiggins:
Are they really them?
We can put that to Okta
And it will tell us.

Logging in with EUA ID via Okta on the frontend now also passed the token to the API, which then checks the token with the Okta `keys` endpoint, and then asks the Okta `userinfo` endpoint for more data about the user (including email address and job codes).

For now, associates EUA IDs with fake users, of which there are still three, AK, AZ, and MA.

Also creates the Django model for AppUser, but doesn't yet populate them via fixtures.

Probably needs a bunch of frontend cleanup in terms of where files are and/or how I have components wrapping other components, and may also need to add more routes to the `SecureRoute` in `wrapSecurity.js`.

Note that visiting `/Profile` will, at the very bottom of the page, display the info we get about the user from Okta.


mblevins-collabralink:
Should the states/territories/roles be hardcoded in the Python? Option lists like that are historically kept outside the code, although I guess that list is pretty stable. Minor issue with everything else we have going on.

frontend/api_postgres/carts/carts_api/views.py
```diff
@@ -331,6 +335,123 @@ def fake_user_data(request, username=None):
...
+    token = token_string.split("Bearer ")[1]
+
+    okta_host = "https://test.idp.idm.cms.gov"
+    okta_root = "oauth2/aus4itu0feyg3RJTK297/v1"
+
+    # Verify that this token really came from the Okta server.
+    # TODO: Add error handling and pulling the URL from config.
+    # u = "https://test.idp.idm.cms.gov/oauth2/aus4itu0feyg3RJTK297/v1/keys"
+    is_valid_token = False
+    pk_url = "/".join([okta_host, okta_root, "keys"])
```
timoballard, line 351: Probably not a high priority now, but this could be made more robust by leveraging the Okta server's discovery endpoint to find this url (`jwks_uri`)

https://test.idp.idm.cms.gov/oauth2/aus4itu0feyg3RJTK297/.well-known/oauth-authorization-server

frontend/api_postgres/carts/carts_api/views.py
```diff
@@ -331,6 +335,123 @@ def fake_user_data(request, username=None):
...
+    okta_auth_res = requests.get(pk_url)
```
timoballard, line 352: Down the road optimization: find out how frequently Okta rolls the keys over and cache them to avoid a round trip here

frontend/api_postgres/carts/carts_api/views.py
```diff
@@ -331,6 +335,123 @@ def fake_user_data(request, username=None):
...
+                is_valid_token = True
+            break
+    except Exception:
+        # TODO: better exception handling here
+        return HttpResponseForbidden()
+
+    if not is_valid_token:
+        return HttpResponseForbidden()
+
+    info_url = "/".join([okta_host, okta_root, "userinfo"])
```
timoballard, line 379: Probably not a high priority now, but this could be made more robust by leveraging the Okta server's openid discovery endpoint to find this url (`userinfo_endpoint`)

https://test.idp.idm.cms.gov/oauth2/aus4itu0feyg3RJTK297/.well-known/openid-configuration

timoballard:
I think my comments could be collapsed into something like: ultimately we may want to encapsulate our interactions with the auth server such that the various urls and keys are fetched from the discovery endpoints and cached

frontend/api_postgres/carts/carts_api/views.py
tonydavydets: agreed.  also, at some point, we might want to store all authentication information  (okta host and root for example) as env variables as opposed to having them in source

cl-adavyd (commented):
I see that the token obtained from okta is being stored in localstorage.  not sure if this is in scope, but storing sensitive information in localstorage leaves them vulnerable to XSS attack.

the preferred option is storing them in cookies with httpOnly option set to true.

https://dev.to/cotter/localstorage-vs-cookies-all-you-need-to-know-about-storing-jwt-tokens-securely-in-the-front-end-15id

just putting it out there.  let me know your thoughts

kaitlin:
@tonydavydets I don't have a technical comment on this, but if we think this will get caught by the pentest, we should definitely resolve now, because we really only have one shot at the pentest. If we have high findings (or certain medium findings) we'll have to delay again because we'll have to fix and re-test before we can go live.

kaitlin:
Just ran this past Eric Hill, and he said @tonydavydets 's solution has come up as a recommendation during pen tests before, and any confirmed XSS examples are automatically assigned a high/critical vulnerability level. So, ultimately it's @StephanieCoppel 's decision but this one is probably worth updating now.

frontend/react/src/wrapSecurity.js
```diff
@@ -34,16 +36,18 @@ const WrappedSecurity = () => {
       </div>
     )
   } else {
+
     return (
       <Router>
         <Security {...config.oidc}>
```
mgwalker, line 42:
```suggestion
        <Security {...config.oidc} tokenManager={{ storage: "cookie" }}>
```

This will prevent Okta from storing the token in localStorage. Okta will still write other stuff there, but it's endpoint and key cacheing, so that should be okay.

tadhg-ohiggins: Trying this out locally now, thanks

jgillis1:
@tadhg-ohiggins Would it be useful if I provided you with the real DEV users and their associated states and roles? Could those just be loaded to our DEV environment?

StephanieCoppel:
> Just ran this past Eric Hill, and he said @tonydavydets 's solution has come up as a recommendation during pen tests before, and any confirmed XSS examples are automatically assigned a high/critical vulnerability level. So, ultimately it's @StephanieCoppel 's decision but this one is probably worth updating now.

yes lets please update now with @tonydavydets solution, want to avoid any hiccups during the pen test. thank you!

frontend/react/src/actions/initial.js
```diff
@@ -38,6 +39,35 @@ export const loadUserThenSections = ({userData}) => {
...
 };
 
+export const secureLoadUserThenSections = ({userData, authState}) => {
+  const xhrURL = `${window._env_.API_POSTGRES_URL}/api/v1/appusers/auth`;
+  const xhrHeaders = {
+      "Authorization" : `Bearer ${authState.accessToken}`
+  };
+
+  return async (dispatch) => {
+    await axios({method: "POST", url: xhrURL, headers: xhrHeaders})
```
mgwalker, line 49: Not for now, but we should extract this header handling out and setup a default axios configuration. That way we don't have to set it everywhere we use axios, and all of our API calls can get it.

frontend/react/src/wrapSecurity.js
```diff
@@ -34,16 +36,18 @@ const WrappedSecurity = () => {
       </div>
     )
   } else {
+
     return (
       <Router>
-        <Security {...config.oidc}>
+        <Security {...config.oidc} tokenManager={{ secure: true, storage: "cookie" }}>
```
mgwalker, line 42: This line should take care of @tonydavydets's comment about local storage.

mgwalker (approved):

Issue #468, opened 2020-08-29
Fixtures for section 3g


cjhskippy:
Ensure that, upon deployment to a new environment, the content for section 3G is loaded.  #356 must be completed first.

A/C:
- [ ] Upon deployment, the questions for section 3G on 2020 reports is populated in the database. 

Note: Work with Tony on CL side to get him up to speed

jgillis1:
ROLLOVER NOTE: The content of 3G has changed several times during Sprint I, the most recent being in the evening on 9/23/2020. As of 8:52 AM 9/24/2020, Andrew needs final confirmation that he is interpreting and implementing the content changes made correctly.

Issues Raised:
- help_drawer

- checkbox about whether or not the State collects data via age ranges which then changes which table the State fills out

Issue #356, opened 2020-08-13
Create JSON for Section 3G


tadhg-ohiggins:
Produce the JSON that will represent section 3g.

A/C:
- [ ] JSON file exists with content for every form element in section 3g.

Related to #339. Blocks #468 

NOTE: Claire - please work with Tony on CL to get him speed

erinzstrenio:
Once the JSON is created, I'll add the google doc to the shared CL folder 

- [ ] @erinzstrenio moves Google doc into CL folder

jgillis1:
ROLLOVER NOTE: The content of 3G has changed several times during Sprint I, the most recent being in the evening on 9/23/2020. As of 8:52 AM 9/24/2020, Andrew needs final confirmation that he is interpreting and implementing the content changes made correctly.

Issues Raised:
- help_drawer

- checkbox about whether or not the State collects data via age ranges which then changes which table the State fills out

erinzstrenio:
@jgillis1 The content hasn't changed significantly during Sprint I. The changes I've made in the doc reflect what we're capable of building - it started as help drawer text then we decided against help drawers for now (not 508 compliant, not a reusable component for other sections, etc), so I reformatted the content to reflect these changes. I removed the checkboxes yesterday per my chat with Greg/Andrew, and reformatted the content to make sense without a checkbox. But the actual words haven't changed!

Issue #315, opened 2020-08-06
Section 3C Build: InputGrid and SynthesizedTable components


cannandev:
AC: 

- [x] Build grid from JSON data
- [x] Hide age breakdown fields when toggle is checked
- [ ] ~~Send total values to store~~
- [x] Update SynthesizedTable with total value
- [x] Render from Question Component

Related to #92 3C and 3I.

cannandev:
blocks 3G build #211

mgwalker:
@tadhg-ohiggins One of the acceptance criteria here is "send total values to store" – is that correct? Do computed values get stored back into the JSON somewhere? If so, do they also get persisted back to the API? I was under the assumption that synthesized values were always computed on the fly.

Or if that's wrong, I'll strike out that checkbox. 😛

tadhg-ohiggins:
This was an ongoing question, but yes, I think the total should get sent back, particularly since (unless I'm mistaking this for another case) sometimes states _only_ enter the totals.
The issue this raises is what happens if there are conflicts between these numbers, and I don't think we ever fully worked that out, other than that anything the user enters should probably take precedence even if it conflicts with the math…

mgwalker:
The case where users could enter totals or have them computed has been removed for now. AFAIK, values are either computed or user-editable, but no longer both.

My biggest concern is any difference in the math implementation causing a difference between what a state user saw when they submitted the form and what a CMS user sees when they're reviewing it. But if that's all handled in the frontend for now, I'm content to leave it out of the JSON for now to avoid the extra complexity.

cjhskippy:
The Q of whether we need to save total values to store went back and forth a few times.  Per conversation yesterday, we decided that was NOT needed for this issue, we're going to close this one and work on related functionality in #718

Issue #211, opened 2020-07-17
Section 3G Build


TaraWilliamsPGH:
As a state user, I want to be able to submit my answers for section 3G.

AC:
- [ ] Section 3G appears in the navigation menu
- [ ] All Form fields for 3G display per [design document](https://docs.google.com/document/d/1R5WJPJlisgvxVqQv4mGhKR6_KXUknHwRvo40a69y_D8/edit).
- [ ] Users can save answers that will be retained when navigating to other parts of the form (e.g. saved in the Redux store)
- [ ] Updates are deployed to the cartsdemo site + pass UAT w/ Stephanie.

Engineering notes:
This will involve these subtasks:
- [ ] #356 - Create JSON for section 3G
- [ ] #468 - Fixtures for 3G
- [ ] New Fieldset set up for Fieldset without fieldset_type
- [ ] New Fieldset set up for Fieldset with fieldset_type = marked
- [ ] New Fieldset set up for Fieldset with fieldset_type = datagrid

TaraWilliamsPGH:
AC:

- [ ] Section is built and rendered

TaraWilliamsPGH:
NOTE: Build team, please review section 3C for structure purposes, however the content will be different.

jgillis1:
@cannandev Should this be assigned to you since you are working on the InputGrid component?

jgillis1:
ROLLOVER NOTE: The content of 3G has changed several times during Sprint I, the most recent being in the evening on 9/23/2020. As of 8:52 AM 9/24/2020, Andrew needs final confirmation that he is interpreting and implementing the content changes made correctly.

Issues Raised:
- help_drawer

- checkbox about whether or not the State collects data via age ranges which then changes which table the State fills out

TaraWilliamsPGH:
Will be opening a new ticket related to a bug that is persistent throughout the application.

PR #635, opened 2020-09-23
Create JSON and Fixtures for Section 3G

andrewadcock:
Create fixtures and JSON for section 3G

Currently, the information-only fieldsets do not utilize the `fieldset_type` parameter. If we think that's bad execution, I propose we add "informational" to the fieldset_type enum. This will better utilize fieldset.js but it makes the code more verbose. The trade-off seems fairly even but I'm open to discussion.

Tickets: #356 #468


jgillis1 (approved):
I added notes to the Build ticket (#211) for components that need to be built out for this section. The JSON matches the updated content document.

cl-adavyd (approved):

jgillis1:

docs/section-schemas/2020-ma-section-3.json
```diff
@@ -3738,8 +3738,501 @@
...
+              "show_if_state_program_type_in": [
+                "separate_chip",
+                "combo"
+              ]
+            },
+            "questions": [
+              {
+                "type": "fieldset",
+                "label": "Note on age groups",
+                "hint": "Children should be in age groups based on their age on September 30th, the end of the federal fiscal year (FFY). For example, if a child turns three years old on September 15th, the child should be included in the “ages 3-5'' group. Even if the child received dental services on September 1st while they were still two years old, all dental services should be counted as their age at the end of the FFY.",
```
mgwalker, line 3758: Nice!

I thought I'd fixed how this displays in #621, but apparently not. I'll get a PR up shortly. The issue is in `CMSLegend` – if the thing being displayed doesn't have an ID, the legend is empty. Fieldsets don't have IDs, so... 😬

mgwalker (approved):
This is soooo much easier to reason about. I should be able to knock out #315 quickly with this structure.

I'll also get the fix up for the fieldset labels. I tried it out locally and it looks like a pretty good stopgap.

PR #646, opened 2020-09-24
Make CMSLegend always display label/hint, even if there is no ID

mgwalker:
- addresses the need to replace the help drawer for 3G


andrewadcock:
Looks like child questions are returning the number followed by NaN instead of the char.

mgwalker:
I love how I checked `if(Number.isNaN(lastHunk))` and if that was true, I then proceeded to just use it right in a string. 😂  Should be fixed now!

andrewadcock (approved):
Looks very nice!

PR #647, opened 2020-09-24
Dev sqlserver

joeaero19:


data/aws/postgres.tf
```diff
@@ -69,7 +69,8 @@ resource "aws_security_group_rule" "vpn" {
 }
 
 resource "aws_db_parameter_group" "db_param_group" {
-  name   = "rds-pg-${terraform.workspace}"
+  name   = "rds-postgres-${terraform.workspace}"
```
dwhitecl, line 72: why is this line changed @joeaero19 ?

dwhitestratiform (commented):
please check one comment in the PR otherwise it looks good to me

data/aws/postgres.tf
joeaero19: It was complaining about the name when I was testing. Ignore this I am correcting it now.

dwhitecl:
no issues from me. waiting for @mdial89f  to review as well

mdial89f (changes requested):
Remove all instances of api sqlserver from the docker-compose.yml and coker-compose.dev.yml

mdial89f (approved):

PR #650, opened 2020-09-24
add NoninteractiveTable

TGriesemer:
Note that the conditional logic seems to be lacking the essential information in context_data so all children are being displayed.


jgillis1 (approved):
Good catch with adding this to the shouldRenderChildren const

cl-adavyd (approved):

jgillis1:

mgwalker (approved):
A suggestion, but take it or leave it. Merge at will!

jgillis1:
I'll take a look at the JSON tomorrow and get that fixed.

tadhg-ohiggins (changes requested):
This doesn't have any code to cover the cases where the percent change is under 10, i.e. to only show the subquestions if it's 10% or more*. There's no logic for that in `context_data` because it needs to be covered by custom code. Unless it's being covered elsewhere, that code should be part of this component.

*: That is, hide the subquestions if the percentage changes are between −10% and 10%.

jgillis1:
If we put it in the component, wouldn't that have to always apply then? So anywhere else the component is used, those parameters would have to apply? I see what you're saying as it cannot be handled in the JSON but this doesn't seem quite right as a solution either. We will pow wow on this tomorrow/Monday and see if we can come up with an appropriate solution.

tadhg-ohiggins:
The most likely route is to tweak the JSON so that the component can read from it. It's possible also that we should move this to a `synthesized_table` and add an action to what that component supports, so that's a consideration.

tadhg-ohiggins:
Having just spoken to Erin, it looks like the threshold is changing from 10% to 3%, and so it's much more likely that a large majority of states are going to have to answer those questions.
That being the case, it seems far less worth it to spend development time on handling this condition, and Erin is fine with changing the question text to reflect it always showing. We should confirm with @StephanieCoppel , but I think the conditional logic for the threshold should be moved to some future enhancement sprint.

jgillis1:
Thank you @tadhg-ohiggins and @erinzstrenio. Changes to the JSON to reflect a new percentage and question wording would be out of scope for this PR, as it applies only to the table component. Assuming Stephanie agrees regarding the conditional logic, @tadhg-ohiggins, could you please approve this PR as the table is displaying properly and the child questions are rendering appropriately.

StephanieCoppel:
@jgillis1 @tadhg-ohiggins i'm good with not including the conditional logic here, so you are good to move forward on this, thanks!!

tadhg-ohiggins (approved):

PR #655, opened 2020-09-24
State image fix

tadhg-ohiggins:
We don't need hostname
To find the state images
And so, remove it.

I was unnecessarily including `full_host` in the image URI for states.

The better fix is to not send it back from the API at all and just determine it from state in the frontend, but we can make that improvement a little later; for now, just make the images work.


mgwalker (approved):
I could have sworn someone else had a PR that did this last week... 🤔  I hope it didn't get lost in a merge somewhere.

PR #658, opened 2020-09-24
Show question IDs of fieldsets with IDs, and allow questions to hide their IDs

### LINKED ISSUE #657: Numbering of Fieldset Children
The fieldset id should be used as the question id and displayed before the fieldset question. Fieldset children (of type datagrid) should only display the question and omit the question number.

Example can be found: http://localhost:81/sections/2020/03/g?dev=dev-ma

AC: 

- [x] Fieldset should show parent question number and hide child question numbers if datagrid

mgwalker:
Also handles the case where a DataGrid contains fieldset children by visually nesting them, but not trying to put them inside the DataGrid cell. 😛 

Closes #657

### DIFF
--- a/frontend/react/src/components/fields/CMSLegend.js
+++ b/frontend/react/src/components/fields/CMSLegend.js
@@ -1,8 +1,9 @@
 import React from "react";
 import PropTypes from "prop-types";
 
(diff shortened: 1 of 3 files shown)


andrewadcock (approved):
Beautiful!

PR #660, opened 2020-09-24
Change section 3C datagrids to remove checkbox and match 3G

### LINKED ISSUE #315: Section 3C Build: InputGrid and SynthesizedTable components

mgwalker:
addresses #315


jgillis1 (changes requested):
- [x] (Optional as it's not really part of this PR scope but is in 3C) Part 3, Question 4 is not editable but should be since it is required for comparison to the Computed value that follows.

- [x] Checkbox is still present on Line 966. Is that correct? If so, "NaNs" is displaying as part of the text.

- [x] Line 1140, the question "How many children had a break in CHIP coverage but we re-enrolled in Medicaid six months later?" is showing as a part of the datagrid for the previous question.

- [x] Line 1253, the question "How many children were no longer enrolled in CHIP but enrolled in Medicaid six months later?" is showing as a part of the datagrid for the previous question.

- [x] Line 1437, the question "How many children had a break in CHIP coverage but were re-enrolled in Medicaid 12 months later?" is showing as a part of the datagrid for the previous question.

- [x] Line 1550, the question "How many children were no longer enrolled in CHIP at the end of the 12-month period, and enrolled in Medicaid instead?" is showing as a part of the datagrid for the previous question.

- [x] Line 1725, the question "How many children had a break in CHIP coverage but were re-enrolled in Medicaid 18 months later?" is showing as a part of the datagrid for the previous question.

- [x] Line 1838, the question "How many children were no longer enrolled in CHIP at the end of the 12-month period, and enrolled in Medicaid instead?" is showing as a part of the datagrid for the previous question.

mgwalker:
- I updated the JSON to make question 4 fillable - might as well, since I was in there anyway!
- I also fixed that checkbox. It was in MA, AK, and the schema, but it I had removed it from AZ. Trying to keep all four files in sync is... a thing!
- for the questions showing within previous data grids, the [design doc](https://docs.google.com/document/d/1pyCyUQPJKVEIKZ3cBTM8ZThxNeLiUzIKPPr0xfImtv4/edit#bookmark=id.91j41jqww7j3) shows that some of them should be nested. Are you seeing this? (I ask because there was a glitch for a while where the child data grid was shoved into a column of the parent data grid and that is... clearly wrong. 😂 )
   screen shot of data grid with another data grid as a child

jgillis1 (approved):

PR #662, opened 2020-09-24
Don't show part numbers if there's just one part in a subsection

### LINKED ISSUE #614: Changes for part titles
As a CMS staff member drafting the CARTS form, I want to be able to put any text into part titles.

Sometimes I want to display text without a part number at all.
- Example 1: On the "basic information" section, just "Welcome" instead of "Part 1: Welcome!"
- Example 2: Section 3a only has one part. It should just say "Program Outreach" and not have separate lines for "Part 1" and "Program Outreach"

Sometimes I want explicit part numbers to be displayed.
- Example 3: Section 1, Part 2 - "Part 2: Separate CHIP Enrollment and Premium Fees") currently displays exactly as desired.

The current frontend code always applies a part number.  It does not currently handle example 1 well but does handle example 2 and example 3.

AC:
- [x] Subheadings for all form parts have the text specified in the [design docs](https://drive.google.com/drive/folders/13HfIL1ye7rg89HbZpx3ICBLJ8UoKqsYM).  Only applies to sections that have been built to date.
  - [x] [Section 6](https://mdctcartsdev.cms.gov/sections/2020/06?dev=dev-ak) should have the heading "Challenges and Accomplishments" where it currently says "Part 1"

mgwalker:
addresses #614


andrewadcock (commented):
This definitely works for hiding part numbers when only one part is present. I don't think this satisfies the requirements of #614. The basic info page has multiple parts but the A/C says "Welcome" shouldn't have a part number.

mgwalker:
Nice catch, thank you!

andrewadcock (approved):

PR #671, opened 2020-09-26
Autogenerate fixtures

tadhg-ohiggins:
~~We don't need fixtures~~
~~If we're generating them~~
~~On each Github push.~~

~~The fixture files are really just build artifacts, and I originally thought having them in version control might make troubleshooting easier, but was wrong.~~
~~Having them as separate version-controlled files adds complexity and has already led to issues with them no longer matching each other.~~
~~I deleted the fixture files and the fixtures directory, added generation of the fixtures to the schema workflow, and added a simple Python script to verify that the contents of the generated Django fixtures match the JSON files in `/docs/section-schemas`.~~

Moved to #681 due to some merge issues.


tadhg-ohiggins:
Converted to draft because I want to think through how best to add this to the development environments once the output is no longer in the repo.

PR #673, opened 2020-09-26
Automated code style checks

mgwalker:
- checks that prettier has been run on all modified files
- checks that all modified Javascript files pass eslint
- updates the prettier commit hook so that it now fails if modified files have not be formatted

To run prettier manually:
- from the `frontend/react` directory, execute `npm run prettier` to have it format your modified files
- from elsewhere, just `npx prettier <filename> -w`

To run eslint manually:
- from the `frontend/react` directory, run `npm install` to make sure eslint and its dependencies are installed (plugins, configs, etc.)
- run `npm run lint` to run eslint on the entire Javascript codebase
- run `npx eslint $(git diff --name-only --relative | xargs)` to run eslint on only the files you've edited

These checks will not currently prevent a merge.

### DIFF
--- a/.github/workflows/style_checks.yml
+++ b/.github/workflows/style_checks.yml
@@ -4,11 +4,22 @@ on: [pull_request]
 
 jobs:
   prettier:
+    name: prettier has been run on compatible files
     runs-on: ubuntu-latest
     container: node:14
     steps:
       - uses: actions/checkout@af513c7a016048ae468971c52ed77d9562c7c819
-      - name: diff from master
-        run: git diff origin/master --
       - name: check that modified files are prettier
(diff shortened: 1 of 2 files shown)


heymatthenry:
[eslint](https://eslint.org/docs/user-guide/command-line-interface#options:~:text=%2D%2Dignore%2Dpath%20path%3A%3AString%20%20%20%20%20Specify%20path%20of%20ignore%20file) and [prettier](https://prettier.io/docs/en/cli.html#--ignore-path) both support an `--ignore-path` flag which I like to set to `.gitignore`, primarily to keep the tools from scanning the node modules dir. I realize since these are only running on changed files this is less of an issue, so it's not a huge deal to omit that. Just making a note of it as something to consider. LGTM otherwise.

heymatthenry (approved):

PR #681, opened 2020-09-28
Autogenerate fixtures as part of build, remove fixture files from repo

tadhg-ohiggins:
We don't need fixtures
If we're generating them
On each Github push.

The fixture files are really just build artifacts. I originally thought having them in version control might make troubleshooting easier, but was wrong.
Having them as separate version-controlled files adds complexity and has already led to issues with them no longer matching each other.
I deleted the fixture files and the fixtures directory, added generation of the fixtures to the schema workflow, and added a simple Python script to verify that the contents of the generated Django fixtures match the JSON files in /docs/section-schemas.


frontend/api_postgres/entrypoint.sh
```diff
@@ -11,6 +11,8 @@ then
     echo "PostgreSQL started"
 fi
 
+python utils/section-schemas/generate_fixtures.py
+python utils/section-schemas/compare_fixtures.py
```
mgwalker, line 15: This script is run in the deployment containers too, right?

mgwalker (approved):
Looks good to me. Just one short question about deploying, but I think it's ready to go.

frontend/api_postgres/entrypoint.sh
tadhg-ohiggins: Yes, it's called by the Dockerfile for both dev and non-dev deploys.

tadhg-ohiggins:
I'll have a look at those merge conflicts.

mgwalker:
Maybe that prettier script can be smarter about files that were deleted, too... 🤔  For another time, though.

PR #684, opened 2020-09-28
User prettier formatter to keep JSON consistent

tadhg-ohiggins:
Prettier JSON
So it's easier to find
Those pesky errors.

With the size and complexity of the JSON files, we should cut out whitespace and formatting changes and ensure it's easy to see real changes in diffs. This adds a Github action to check formatting of the JSON files against what `prettier` expects.

Technically not dependent on #681, #682, and #683, but adding it before those are added will mean having to do a bunch of redundant JSON editing, so we shouldn't do that. Compare against tadhg/autogenerate-state-json: https://github.com/18F/cms-carts-seds/compare/tadhg/autogenerate-fixtures...tadhg/prettier-json


mgwalker:
@tadhg-ohiggins Is this PR still relevant?

mgwalker:
Obsolete by prior PRs

PR #686, opened 2020-09-28
Ignore deleted files for prettier

mgwalker:
There's not really any good reason to complain about deleted files. Also ignore generated fixtures, so they don't accidentally get checked in again.


tadhg-ohiggins (approved):
Looks good.

PR #688, opened 2020-09-28
Remove initial.json fallback

mgwalker:
The `initial.json` fallback was helpful when we were still getting the API up and running, but now that it's more reliable, we should remove the fallback. It's now creating a false sense of confidence by allowing the site to load even when the API is unreachable for some reason. This PR will make the site fail to render if it can't get data from the API, making it faster for us to notice when the API is down.

The diff is bigger than necessary because I moved some stuff around to satisfy eslint.


frontend/react/src/actions/initial.js
```diff
@@ -49,36 +71,18 @@ export const secureLoadUserThenSections = ({ userData, authState }) => {
         dispatch(getStateData(res.data));
         dispatch(getUserData(res.data.currentUser));
       })
-      .catch((res) => {
+      .catch((err) => {
         /*
```
heymatthenry, line 75: I would just remove this block comment since it's duplicative of the one above and the comment below better reflects what's happening right here.

frontend/react/src/actions/initial.js
```diff
@@ -18,24 +36,28 @@ export const loadUserThenSections = ({ userData }) => {
         dispatch(getStateData(res.data));
         dispatch(getUserData(res.data.currentUser));
       })
-      .catch((res) => {
+      .catch((err) => {
```
heymatthenry, line 39: 👍 to this and renaming `userData` to `fakeUser`

heymatthenry (approved):

jgillis1 (approved):

PR #690, opened 2020-09-28
Add Python style check

mgwalker:
Runs a [black](https://github.com/psf/black) check on modified Python files.


tadhg-ohiggins:
Only if we set line length to 79

tadhg-ohiggins (approved):
79 is the magic number

PR #691, opened 2020-09-28
Comment out write_state_json to see if that changes the build failure

tadhg-ohiggins:
Without state JSON
Will everything else succeed?
Joyous bug-finding.


timoballard (approved):
:amaze:

Issue #206, opened 2020-07-17
User Roles within Postgres Database


TaraWilliamsPGH:
![user roles.PNG](https://images.zenhubusercontent.com/5f2069da63dfc750bb46f86f/7f7dd437-34d4-4175-9c2f-6ccf123acab3)Setting up permissions, associations to states, etc.

AC:
- [x] Define various roles (Ex. state user; cms user [PO has same); administrator(DSCP, Help Desk and Application Team)]
  - [x] User roles defined in Django model.
  - [x] ~~Fixtures loaded so that a wider range of fake users are available via Django.~~ Not via fixtures, but yes.
  - [x] Okta auth in dev/staging environments assigns dev users to one of those roles rather than just to a state user role.
  - [x] Fake auth allows bypassing and specifying roles other than state user.
- [x] User profiles include identifying information (name, email used for CARTS), association to a state if applicable, and assigned role.
  - [x] Okta auth checks for job codes to verify that user is allowed to have the role that the DB thinks they should have.
  - [x] After the assignment to dev users described above works, check against and make available the data the DB has about the user.
- [x] CO and admin users can see various state information without having to log in again.
  - [x] Endpoints set up for per-state views.
  - [x] Primarily a matter of frontend routing, and of dissociating determining the user from loading the state data.
  - [x] Endpoints support querying by status.
  - [x] Backend endpoints have auth added so that state users cannot make backend requests to other states' data.

- [ ] ~~Define endpoint for front end to read user data~~ (moved to #481)

QA Testing Notes:
1. If I am a state user I cannot see the same level of detail as a CMS user (example)

tadhg-ohiggins:
Addressed by #789 and #794.

tadhg-ohiggins:
#789 is now in mdctcartsdev.cms.gov
EUA users with the `CARTS_Group_Dev` will be assigned a role that has no state but lets them view the three states we have data for, e.g. https://mdctcartsdev.cms.gov/views/sections/AK/2020/0 https://mdctcartsdev.cms.gov/views/sections/AZ/2020/0 and https://mdctcartsdev.cms.gov/views/sections/MA/2020/0 will show Section 0 for the three states without requiring logging in again.

There is currently no frontend UI for this.

There is also a `?dev=dev-admin` role that allows seeing all of the state info; this role needs to be renamed but is the CMS/DSCP role.

https://mdctcartsdev.cms.gov/userinfo shows the user info.

[...]

PR #696, opened 2020-09-29
Add program_type and program_names to State model

tadhg-ohiggins:
Program type and names
Belong to states not users.
Add them to the states.

- Add `program_type` and `program_names` to `State` model.
- Add list of `program_type` options.
- Add `State`, `ACS`, `FMAP`, and `AppUser` to models visible in Django admin.
- Use `state-fixture-data.csv` for populating the state fixtures with the new program_type and program_names fields.
- Remove American Samoa, Guam, the Northern Mariana Islands, Puerto Rico, and the Virgin Islands from the fixture data, because none of them report to CARTS.

This is related to #206 because separating what's associated with states versus users is necessary before developing the models/endpoints, and users are going to be associated with states and pull program type and program names from the states, so we need to populate that info.


tadhg-ohiggins:
FWIW I did test this against a running database with prior data in it rather than a completely new database, and the migrations and fixtures worked fine, but that's not a guarantee that that'll work in staging…

tadhg-ohiggins:
@kaitlin @erinzstrenio @StephanieCoppel I removed American Samoa, Guam, the Northern Mariana Islands, Puerto Rico, and the Virgin Islands from the FMAP and state data, let me know if we need to add them back for some reason.

StephanieCoppel:
> @kaitlin @erinzstrenio @StephanieCoppel I removed American Samoa, Guam, the Northern Mariana Islands, Puerto Rico, and the Virgin Islands from the FMAP and state data, let me know if we need to add them back for some reason.

@tadhg-ohiggins we do not need to add those back, thanks for removing those. thanks!

frontend/api_postgres/carts/carts_api/models.py
```diff
@@ -40,30 +43,64 @@ class State(models.Model):
...
+    number_uninsured = models.IntegerField(
+        help_text="the number of uninsured children under 200% FPL"
+    )
+    number_uninsured_moe = models.IntegerField(
+        help_text="the margin of error for the number of uninsured children"
+    )
+    # help_text has exactly five spaces between “uninsured” and “children” in
+    # order to avoid triggering a database migration for such a minor issue.
+    percent_uninsured = models.DecimalField(
+        help_text="percentage of uninsured     children",
```
mgwalker, line 94: The result of reformatting the code, I'm guessing? Does the backslash preserve whitespace?

mgwalker (approved):
Looks fine to me, but having @timoballard's eyes on `generate_fixtures` wouldn't be the worst idea.

timoballard (approved):

frontend/api_postgres/carts/carts_api/model_utils.py
```diff
@@ -61,6 +67,6 @@
 )
 
 USER_ROLES = (
-    ("state_user", "State User"),
     ("co_user", "Central Office User"),
+    ("state_user", "State User"),
```
timoballard, line 71: this reordering results in a new migration being generated - I think we should either back this change out or commit the migration

timoballard (changes requested):

frontend/api_postgres/carts/carts_api/model_utils.py
tadhg-ohiggins: I'll back it out.

tadhg-ohiggins:
@timoballard Could you verify that this version doesn't trigger a migration? If it's good to go after that, feel free to merge it in.

jgillis1:
As a note, please be aware that states can change/add/remove programs year over year. So the 2020 program name and type could be different in 2022 (i.e. the 2020 program is discontinued and replaced with a new program). Does that need to be accounted for in this model or would it be handled in the database only?

timoballard:
@tadhg-ohiggins - just checked, no migrations now 👍

tadhg-ohiggins:
@jgillis1 Yes, I was thinking about this—do we need to track state historical program type, or would it be enough to have that info in their answers for previous years? So if MA is a combo state in 2020 but a separate_chip state in 2021, would it be enough for their CARTS report itself to have that data but their state info in the DB say they're separate_chip? I was assuming that having the info in the reports would be sufficient. @StephanieCoppel any thoughts on this?

jgillis1:
We probably don't (though the current system does) but Program Type drives
behavior in the app in terms of what is shown and not shown in some
sections. You're saying that the form then would be the source of truth for
program type and name rather than what you are loading into the state
model? And, follow up question, if that is so, what is the purpose of
adding them to the state model? (Just curious there)


*Jenna Gillis*
Full Stack Developer, ACME Dev Team Lead
CollabraLink Technologies, Inc.
jgillis@collabralink.com
*C:* 443.807-7580

jgillis1:
@tadhg-ohiggins We probably don't need it historically (though the current system does) but Program Type drives behavior in the app in terms of what is shown and not shown in some sections. You're saying that the form then would be the source of truth for program type and name rather than what you are loading into the state model? And, follow up question, if that is so, what is the purpose of adding them to the state model? (Just curious there)

tadhg-ohiggins:
The state model would be the source of truth for active forms. So if MA is combo in 2020 that would be shown in the state model. But if MA had been separate_chip in any prior year, that would only be captured in the MA report for that year.

timoballard (approved):

jgillis1:
@tadhg-ohiggins Could you add comments to the model RE: the above so the set up is easy to find and reference in the future?

PR #705, opened 2020-09-29
Save form data to database

### LINKED ISSUE #475: Form saves to database

timoballard:
Implements form saving per #475, without authentication or authorization checks which are being implemented separately.

- [x] data validation
- [ ] authentication check
- [ ] authorization check


frontend/deployDev.sh
```diff
@@ -1,8 +1,6 @@
 #!/bin/bash
 
-sh destroy.sh
+sh destroyDev.sh
 
-docker-compose -f docker-compose.dev.yml build
```
tonydavydets, line 5: was changing the shell scripts needed for this task as well?

everything else looks good

timoballard: :busted: the shell scripts are pork, changed them because deployDev.sh is currently broken and causes duplicate section data to be seeded into the database, which caused me some heartache while working on this issue.

tadhg-ohiggins: Changing them to refer to `destroyDev` rather than `destroy` is necessary, as is removing the stray reference to `local-additional.sh`.
Changing from separate `up` and `build` commands and away from running as a background process is a matter of preference.

tonydavydets: yup, that is 100% true.  ...could we keep them out this PR though (not to mix functionality / for better trackability).

but you are absolutely right, it was on the list of things to do

tadhg-ohiggins (approved):

frontend/deployDev.sh
tonydavydets: @tadhg-ohiggins I agree, it's just not a part of this task.  I want to avoid confusion / make sure others can follow the thread of what is happening

frontend/api_postgres/carts/carts_api/views.py
```diff
@@ -110,7 +110,21 @@ def sections_by_year_and_state(request, year, state):
...
+            section_id = entry['contents']['section']['id']
+            section_state = entry['contents']['section']['state']
+
+            section = Section.objects.get(contents__section__id=section_id,
+                                          contents__section__state=section_state.upper())
+
+            section.contents = entry['contents']
+            section.save()
+
+    except Section.DoesNotExist:
```
mgwalker, line 125: Is it possible to wrap this in a transaction? I'm interested in the case where there are 3 sections and the 2nd one is invalid. Then the first one is persisted, the invalid one is rejected, and the third valid one is skipped altogether.

cl-adavyd (approved):

tadhg-ohiggins (approved):

PR #713, opened 2020-09-29
Update JSON action

mgwalker:
The JSON action only runs if the JSON files have changed. They moved recently, but the action missed the memo. This uh... delivers the memo.


tadhg-ohiggins (approved):
The memo was in the mail

PR #715, opened 2020-09-30
Add Python linting/formatting configuration

tadhg-ohiggins:
Our Python code style
Should at least agree on these;
Put them in one place.

Config options for `black`, `flake8`, and `pylint`.
None of these are enforced yet, but putting the configs here helps standardize.
The pylint settings are my personal preference, although if they weren't rather annoying I wouldn't have disabled them. `import-error` should definitely be disabled, because otherwise editing in your local environment will cause pylint to complain if a library isn't available—even if the library is available in the Docker container, which is what we care about.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+# Not ignoring these will clash with black
+ignore = E203, W503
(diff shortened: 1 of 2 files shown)


mgwalker (approved):
I have no strong feelings one way or the other, but consistency will be nifty.

Issue #718, opened 2020-09-30
Age range calculations: Do not allow totals to be inconsistent with individual age brackets


cjhskippy:
In sections 3C and 3G, users are asked to provide two things:
- Data for children broken down by age groups (0-3, 4-6, etc.)
- Total data for children of all ages (0-16 in 3c; 0-18 in 3g)

Review of the current design leaves us worried that users will enter inconsistent data or be confused by the form structure.

In order to collect better data, reorganize questions in 3C and 3G to:
- [x] Add a question that verifies whether the state has data broken down by age groups (vs the total for all ages)
- [x] If the user answers yes, they do have data for individual age groups, show the form fields for each age group 
- [x] If the user answers no, they don't have data for age groups, only show an input field for the total
  - [x] Automatically calculate the totals based on the user's individual age group entries.

Related in 3C:
- [x] Update child questions (currently appearing as questions 3f, 4f, etc) to be same level as parent questions, and update question numbers/ids accordingly. 
- [x] Questions pertaining to `12 months later` and `18 months later` should be inactive for states to fill out next year

Content is updated to reflect these changes in the [template doc](https://docs.google.com/document/d/1pyCyUQPJKVEIKZ3cBTM8ZThxNeLiUzIKPPr0xfImtv4/edit?ts=5f6cea35#).

mgwalker:
Making a note for myself: [here's an older commit](https://github.com/18F/cms-carts-seds/blob/612fd1d9435eb44763555562407f1e853fcb8d7d/docs/section-schemas/2020-ak-section-3.json#L941) showing how the checkbox was implemented, for reference.

PR #726, opened 2020-09-30
Toggle between age breakdowns with calculated total and just the total

mgwalker:
Addresses #718 

- puts back the "age breakdown" checkbox
- the checkbox toggles a single input for all ages OR a data grid for the age breakdown that includes a computed total
- updates all the text content to match the document
- disables questions for the future (questions 11-19)
- adds a placeholder for Section 3C part 4, and moves the old part 4 to part 5
- adds section 3C part 6 (basically a repeat of part 5, so it was mostly copy+paste)
- gives the same treatment to section 3G
- adds a helper `<Text>` component that handles breaking big JSON text into multiple lines or paragraphs, so we don't have that logic in multiple places, and updates the `Subsection` and `CMSLegend` components to use the `Text` helper

### DIFF
--- a/frontend/react/src/components/fields/DataGrid.js
+++ b/frontend/react/src/components/fields/DataGrid.js
@@ -11,11 +11,7 @@ export const DataGrid = ({ question }) => {
   return (
     <div className={`ds-l-row input-grid__group ${rowStyle}`}>
       {question.questions.map((input) => (
-        <div
-          className={
-            input.type === "fieldset" ? "ds-c-choice__checkedChild" : "ds-l-col"
-          }
-        >
+        <div className="ds-l-col">
           <Question hideNumber={input.type !== "fieldset"} question={input} />
         </div>
       ))}
--- a/frontend/react/src/components/fields/Question.js
+++ b/frontend/react/src/components/fields/Question.js
@@ -71,8 +71,7 @@ const Question = ({ hideNumber, question, setAnswer, ...props }) => {
 
   const shouldRenderChildren =
     (question.type !== "fieldset" ||
-      (question.type === "fieldset" &&
-        question.fieldset_type === "noninteractive_table")) &&
+      question.fieldset_type === "noninteractive_table") &&
     question.type !== "objectives" &&
     question.type !== "radio" &&
     question.type !== "repeatables" &&
--- a/frontend/react/src/components/layout/Part.js
+++ b/frontend/react/src/components/layout/Part.js
@@ -3,6 +3,7 @@ import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import { Alert } from "@cmsgov/design-system-core";
 
+import Text from "./Text";
 import { selectFragment } from "../../store/formData";
 import Question from "../fields/Question";
 import { selectQuestionsForPart } from "../../store/selectors";
@@ -37,7 +38,7 @@ const Part = ({
   if (show) {
     innards = (
       <>
-        {text ? <p>{text}</p> : <></>}
+        {text ? <Text>{text}</Text> : null}
 
         {questions.map((question) => (
           <Question key={question.id} question={question} />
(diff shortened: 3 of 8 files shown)


frontend/react/src/components/fields/Question.js
```diff
@@ -71,8 +71,7 @@ const Question = ({ hideNumber, question, setAnswer, ...props }) => {
 
   const shouldRenderChildren =
     (question.type !== "fieldset" ||
-      (question.type === "fieldset" &&
-        question.fieldset_type === "noninteractive_table")) &&
+      question.fieldset_type === "noninteractive_table") &&
```
mgwalker, line 74: `!x || (x && y)` is algebraically equivalent to `!x || y`, so I simplified.

frontend/react/src/components/fields/DataGrid.js
```diff
@@ -11,11 +11,7 @@ export const DataGrid = ({ question }) => {
   return (
     <div className={`ds-l-row input-grid__group ${rowStyle}`}>
       {question.questions.map((input) => (
-        <div
-          className={
-            input.type === "fieldset" ? "ds-c-choice__checkedChild" : "ds-l-col"
-          }
-        >
+        <div className="ds-l-col">
```
mgwalker, line 14: DataGrids no longer have DataGrids within them, so this special classing isn't necessary anymore. 🎉

frontend/react/src/components/layout/Part.js
```diff
@@ -32,12 +32,12 @@ const Part = ({
 }) => {
   let innards = null;
 
-  const [_, section] = partId.split("-");
```
mgwalker, line 35: I accidentally changed this file at one point, which then made me go ahead and fix the eslint complaints.

frontend/react/src/util/shouldDisplay.js
```diff
@@ -4,25 +4,25 @@ import jsonpath from "./jsonpath";
  * This function determines whether a radio's conditional subquestion should hide
  * @function hideIf
  * @param {object} state - The application state from redux, the object required for jsonpath to query
- * @param {object} hideIf - the hide_if object from a question's context_data
```
mgwalker, line 7: Also touched this one and ended up linting it.

frontend/react/src/components/fields/Question.js
heymatthenry: Sad to say i had to write out a truth table to convince myself of this

jgillis1:
Do we have a ticket or documentation that addresses how/when to enable the questions for the future RE: "disables questions for the future (questions 11-19)"

frontend/react/src/components/layout/Text.js
```diff
@@ -0,0 +1,24 @@
+import React, { useMemo } from "react";
+
+const Text = ({ children }) => {
+  return useMemo(() => {
+    if (children.split) {
+      return children.split("\n\n").map((paragraph, index) => {
```
jgillis1, line 6: I think in other places we had originally started with just \n as the new line character in the JSON. Did you verify there were no other places in the JSON that needed to be changed due to switching to \n\n?

jgillis1 (approved):

frontend/react/src/components/layout/Text.js
mgwalker: We did! The way we did it was with `<div>` though, so they were always just single spaced since an empty div doesn't take up any vertical space. So what I did here was `\n` still wraps the content with a `<div>` for a single line break, but `\n\n` wraps the content with a `<p>` so it gets two lines.

I checked around and it looked like everything behaved the same, except for a couple of places where `\n\n` had been treated as a single line but was now treated as two lines, which seemed correct.

mgwalker:
@erinzstrenio Good question above from @jgillis1 about when the "next year" questions get turned on. Do you know if we have it documented/created an issue for that?

frontend/react/src/components/fields/Question.js
mgwalker: video of a man being shy

I did too.

erinzstrenio:
I don't believe we have an issue for this yet! 
Note: this only applies to Section 3C, Parts 5 and 6, questions 10-20

Issue #727, opened 2020-10-01
Synthesized Tables with Unknown FMAP data should display "Not Available"


andrewadcock:
As a user, if FMAP data is unknown in a synthesized table, data returned should display "Not Available". 

A/C
- [ ] In section 5, part 2, table 3, `FMAP` row: Data is displayed for FY2020, FY2021, and FY2022. If these values are unavailable, the cell should display "Not Available".
- [ ] In section 5, part 2, table 3, `Federal Share` row: Data is the product of several fields multiplied by FMAP data. If the FMAP data is unavailable the cell should display "Not Available".

heymatthenry:
#734 Referenced above added the functionality to have cells with missing FMAP data display as "Not Available" and it *should* work for cells that perform computations on missing FMAP data too. Does that unblock you, @andrewadcock?

PR #734, opened 2020-10-01
Allow FMAP data to be targets for synthesized value calculation

### LINKED ISSUE #733: Allow FMAP data to be used in synthesized value calculation
As a user, I want to the data I enter to be automatically combined with FMAP data (when available).

- [x] Add an additional type of target that pulls in the FMAP data from Redux.
- [x] These targets should behave like other jsonpath targets in calculations
- [x] If the result of a calculation is `NaN` (such as when FMAP data are unavailable, the cell should display "Not available").

heymatthenry:
This PR revises the implementation of the FMAP lookup to allow those values to behave like other targets and to be used in synthesized value calculations.

As shown here, if the data for the specified FY are unavailable, the value will be displayed as "Not Available". 
image
The value will remain `NaN` if used in other calculations resulting in further `NaN`s down the chain.

This should hopefully support further work on #727 

closes #733

### DIFF
--- a/docs/section-schemas/backend-json-structure-documentation.rst
+++ b/docs/section-schemas/backend-json-structure-documentation.rst
@@ -646,18 +646,6 @@ Example of using ``rpn``:
 
 The above would display a value equal to ``(((q1 + q2) - q3) / q4) * 3``
 
-One additional type of synthesized value that behaves a bit differently from those that perform calculations is the ``lookupFmapFy`` action. This pulls the enhanced FMAP for the current state. Just add the ``lookupFmapFy`` property to a synthesized table cell and specify the fiscal year as the value. This example shows one such cell in a synthesized table:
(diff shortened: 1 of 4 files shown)


mgwalker (approved):
LGTM.

It's got me thinking about how we might further extend the JSON. E.g., for synthesized values, maybe there's another property called something like `fallback_value` where we can specify `Not available`, since that's not going to be appropriate in _every_ case. Erin and I were talking about this yesterday because in some other tables, it also displays `NaN` but it should probably say something like `invalid value` or something instead because the user hasn't entered valid data yet.

It also reminded me that we should probably add data types (e.g., `currency`, `percent`, etc.) to the synthesized values so we can format them properly. I'll open an issue for that one so it doesn't have to rely on just floating around in my brain.

PR #737, opened 2020-10-02
State permissions/permissions groups

tadhg-ohiggins:
Permissions and groups
Soon we will have real users.
This gets us closer.

- Changes both the bypass and Okta dev paths to return information that pulls from the actual DB states and permission groups.
- Adds a perfunctory `/userinfo` page to show what data we're getting from the auth endpoint about the user.
- Adds step for adding the permissions and groups to the Python build steps.

Local testing didn't hit any database snags, but we'll have to see what happens when we push this to `master` in case there are more DB loading issues.

Still related to #206.


tadhg-ohiggins:
Ignoring `black`'s complaints about `views.py` because I don't want to reformat the entire file as that'll make this PR harder to follow.
Ignoring some eslint complaints about prop types because they really don't seem worth adding right now.

frontend/api_postgres/carts/carts_api/models.py
```diff
@@ -56,6 +47,13 @@ class State(models.Model):
     program_names = ArrayField(models.CharField(max_length=64), default=list)
 
 
+class AppUser(models.Model):
+    state = models.ForeignKey(State, on_delete=models.CASCADE)
```
timoballard, line 51: What value does this take for central office users?

tadhg-ohiggins: That is a good point—do you think `null=True` here (which I forgot to add) is enough to handle this, or do you think we also need to account for multiple states in some way? I was hoping the permissions/groups would be enough and that we would have this field purely for getting information for the `state_user` case.

timoballard: I think multiple states is fully handled by the permissions/groups. 

I guess that raises a different question: functionally, does this field do anything for us? Do we ultimately care which state a user belongs to, or does everything boil down to their permissions?

tadhg-ohiggins: We care because we're going to be looking up the same state data (full state name, and type of program primarily) for the user over and over, so it seemed like a tight coupling made sense. I'm open to other suggestions.

timoballard: 🤔  I think I agree

So this `state` sort of represents something conceptually different from permissions, like "default state info for this user", even though in practice they're related (i.e. if `user.state==WI` then they would also have `WI` permissions)

timoballard: if ☝️  is the case, then I think making it nullable makes sense

tadhg-ohiggins: That's how I've been thinking about it, yes.

tadhg-ohiggins:
Converting to draft until #705 is approved, since it'll need to be merged into this branch.

frontend/api_postgres/carts/carts_api/views.py
```diff
@@ -425,48 +406,37 @@ def authenticate_user(request):
...
+    # they're associated with the above state and grab the group for that
+    # state's permissions:
+
+    auth_group = Group.objects.get(
+        name=f"Users who can edit and view {fake_user_key} sections"
+    )
+    state = State.objects.get(code=fake_user_key)
+    assert auth_group and state
+    program_names = ", ".join(state.program_names)
+    program_name_text = f"{state.code.upper} {program_names}"
```
timoballard, line 419: `upper()`

timoballard (changes requested):

frontend/api_postgres/carts/carts_api/views.py
tadhg-ohiggins: Good catch, thanks!

timoballard (approved):

Issue #739, opened 2020-10-02
Section 3b Q4 shows conditional questions for all responses


jgillis1:
The conditional questions for Section 3b Question 4 (4a-4d) should only be shown on a "Yes" response. A "No" or "N/A" response should not show the conditional questions.

Additionally, It looks like Question 4 should actually only be shown IF the State has a Separate CHIP (SCHIP) program.

AC:

- [x] 4a-4d only show for a "Yes" response to Question 4

- [x] Question 4 shows if a State is NOT a Medicaid Expansion CHIP program (MCHIP)

jgillis1:
Relates to #697

Issue #738, opened 2020-10-02
Allow synthesized values to have a data type, so they can be visually formatted


mgwalker:
It would be useful if we could specify in the Javascript what a synthesized value's type is (e.g., `currency`, `percent`, etc.) so that we can format it appropriately.

mgwalker:
Related to #710

PR #748, opened 2020-10-02
Skip text bug 

### LINKED ISSUE #654: Skip Text within Repeatable Not Displaying
Example: 3i, Part 2, if Question 2 is answered "No", a skip text should display.

AC:

- [ ] skip_text within a repeatable is displayed

alewoo28:
**Overview:**
Fixed broken instances of skip_text. Issue #654 which were largely JSON issues. 
The specific case mentioned in the issue (3i, Part 2, if Question 2) was the result of `skip_text` being in the wrong place. 

**Files changed (2):**

**docs/section-schemas/backend-json-structure-documentation.rst**
The issue I was finding with skip_text is that it was often just in the wrong place. Skip text is either being handled as a question or as a part. Fieldsets with skip_text often had skip_text in the place it needs to be for parts, not questions. My intention with editing the docs is so that anyone building JSON for fieldsets after this knows where skip_text needs to be in order to be evaluated. 

**frontend/api_postgres/utils/section-schemas/backend-json-section-3.json**
section 3D:  
the conditional logic and skip text for section 3d were incorrect. I put questions 2-8 into a fieldset with conditional logic so that the skip_text of question 7 displays and hides as its supposed to. 
section 3I: 
Moved skip text into `conditional_display` object. 

Other places where skip_text isnt working involve larger problems and they have separate tickets. 
2020-02-a-01-01: Is related to non-interactive tables not being complete. There just needs to be a context_data object when this is built. (issue #738) 
2020-03-b-01-04-a: This question is being treated like a part. Needs a conditional_display object. The conditional logic on this one is a bit funky, it relies on both program type and conditional logic (issue #739).


jgillis1:
A testing note, Section 3d is only available for AZ in our group of test states so if you are auto-directed to a state other than AZ, you will need to add ?dev=dev-AZ to the end of the URL.

frontend/api_postgres/utils/section-schemas/backend-json-section-3.json
```diff
@@ -1968,202 +1876,234 @@
...
                       "conditional_display": {
+                        "skip_text": "Hidden based on your answer to the cost sharing questions in Section 1 Part 4.",
                         "type": "conditional_display",
-                        "comment": "Interactive: Hide if 2020-03-d-01-06 is no or unanswered; noninteractive: hide if that's no.",
+                        "comment": "Interactive: Hide if 2020-01-a-04-08",
                         "hide_if": {
-                          "target": "$..*[?(@.id=='2020-03-d-01-06')].answer.entry",
+                          "target": "$..*[?(@.id=='2020-01-a-04-08')].answer.entry",
                           "values": {
                             "interactive": [null, "no"],
```
jgillis1, line 2091: I'm on the fence with this. Should Question 7 display if there is a null value for Section 1, Part 4, Question 8?

One would think Section 1 would be filled out before Section 3D, but what if it's not for some reason? We don't want our end user to miss answering a question they should have as a result. I'm erring on  the side of removing null as a hide_if condition though personally.

Think about it but I'm going to approve the PR because functionally this is working properly.

jgillis1 (approved):

jgillis1:
Very nice write-up Alexis. Thank you for being so detailed.

frontend/api_postgres/utils/section-schemas/backend-json-section-3.json
mblevins-collabralink: I concur - since the pages are random access, there's nothing to *stop* a user from doing the sections out of order. It's probably an edge case, but not that far out to the edge.

mblevins-collabralink (approved):
This is how a writeup should be done. Very nice and informative.

I'm approving even though I still have my doubts about the question 7 logic, it does meet the AC.

frontend/api_postgres/utils/section-schemas/backend-json-section-3.json
alewoo28: Thats a really good point, I hadnt thought of that. I feel like that makes a good case for there being some final check before a form is submitted. Something that either confirms all questions have been answered or brings to their attention what has not been answered. 

Ty for bringing it up I think thats a great use case to have in mind

alewoo28: Updated! The null option has been removed

jgillis1:
QA review is completed and this is ready for 18F review.

mgwalker (approved):
This looks good to me and an A+ writeup.

video of a guy giving thumbs up in front of fireworks

PR #750, opened 2020-10-02
American community survey

andrewadcock:
Decided to follow the same pattern as the lookupFMAP action since this information only needs to be on one page. This has an added benefit of being easily manipulated for the front-end; flexibility with table headers, ordering, etc. 

JSON for section 2 was modified to show the previous 5 years of data. The percent change non-interactive table and its child, were moved because synthesized table cannot handle child questions 

Updated _form.scss to add spacing around tables.


frontend/api_postgres/utils/section-schemas/backend-json-section-2.json
```diff
@@ -70,77 +60,132 @@
...
                   ]
                 },
+                "questions": []
+              },
+              {
+                "type": "fieldset",
+                "fieldset_type": "noninteractive_table",
+                "fieldset_info": {
+                  "headers": ["Percent change between 2018 and 2019"],
+                  "rows": [[-1.7]]
```
jgillis1, line 179: Should this be a synthesized value instead? Or some calculation instead of hard coded. If you think we need a new ticket to address this specifically, let me know.

jgillis1 (approved):

mblevins-collabralink (approved):

cl-adavyd (approved):

jgillis1 (approved):

jgillis1:

frontend/react/src/util/synthesize.js
```diff
@@ -133,11 +133,96 @@ const lookupFMAP = (state, fy) => {
...
 };
 
+/**
+ * Retrieve acs_set from state and return for individual state.
+ *
+ * @param {string} state
+ * @param {array} args
+ * @returns {string}
+ */
+const lookupAcs = (state, args) => {
```
mgwalker, line 143: Go ahead and destructure `args` here so it's clearer later on what they're doing:

```js
const lookupAcs = (state, [ffy, acsProperty]) => {
```

frontend/react/src/util/synthesize.js
```diff
@@ -133,11 +133,96 @@ const lookupFMAP = (state, fy) => {
...
+};
+
+/**
+ * Retrieve acs_set from state and return percentage change for 2 given years.
+ *
+ * @param {string} state
+ * @param {array} args
+ * @returns {(string|float)}
+ */
+const compareACS = (state, args) => {
```
mgwalker, line 173: Same story

```js
const compareAcs = (state, [ffy, acsProperty]) => {
```

mgwalker (changes requested):
Looks nice and clean to me! A pair of small change requests, and one open question (which I will happily defer to your judgement on, especially since "leave it alone" is less work).

mgwalker (approved):

PR #760, opened 2020-10-05
Don't apply focus ring on forms during a click

### LINKED ISSUE #707: Black Outline
Section 3, Part I, "Part 2", Question 2 **and **Section 4, Question 2 & 8:****

When selecting the radio buttons "Yes" or "No" a black box outlines all of the sections briefly. This is not consistent with behaviors in the rest of the form.

heymatthenry:
The issue here is the browser stylesheet's focus style being applied briefly when clicking on something in the form that isn't an input:

before

This PR fixes the issue by disabling the focus ring during the click:

after

One small caveat that we shouldn't make changes to focus states that are more extensive than this, for accessibility purposes.

fixes #707

### DIFF
--- a/frontend/react/src/scss/_form.scss
+++ b/frontend/react/src/scss/_form.scss
@@ -1,7 +1,8 @@
 /**********************
   FORM
 **********************/
-.errors, .error {
+.errors,
+.error {
   color: #e31c3d;
 }
 
@@ -15,6 +16,10 @@
   button {
     margin-right: ($padding * 3);
   }
+}
+
+[role="region"]:active:focus {
+  outline: none;
 }
 
 .accordion-title {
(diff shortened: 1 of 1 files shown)


frontend/react/src/scss/_form.scss
```diff
@@ -1,7 +1,8 @@
 /**********************
   FORM
 **********************/
-.errors, .error {
+.errors,
```
heymatthenry, line 4: Just a formatting fix from prettier

frontend/react/src/scss/_form.scss
```diff
@@ -221,7 +224,7 @@
 .autosave {
   float: left;
   margin: 64px 0 16px;
-  text-align: center;  
+  text-align: center;
```
heymatthenry, line 227: Just whitespace

mgwalker (approved):

Issue #697, opened 2020-09-29
Allow questions to be hidden by program type


andrewadcock:
Certain questions need to be shown/hidden based on program type. Currently, this only works for parts and not individual questions.

This can be seen by using `show_if_state_program_type_in` in section 5 (near the top).
`Combine your costs for both Medicaid Expansion CHIP and Separate CHIP programs into one budget.` should only be shown for combo states.

A/C:
- [ ] Show/Hide questions based on state program type

PR #780, opened 2020-10-07
Allow questions to be hidden/shown by program_type

andrewadcock:
Related to #697 and #739

**shouldDisplay.js:** now allows for checks against `show_if_state_program_type_in`. As a result the check for acceptable context and context properties was updated

**Selectors.js:** Update to show skip_text in the format of `show_if_state_program_type_in` which puts the skip_text in a different location than other hide_if* actions. This file has lots of lint errors, most were addressed, however, two were hidden from the linter because they require large workarounds that seem out of scope for this ticket.

**backend-json-section-3.json:** Updates to the hide_if to hide questions 4a - 4d if q4 is answered no, n/a, null.


jgillis1 (approved):

mblevins-collabralink (approved):

jgillis1:

frontend/react/src/store/selectors.js
```diff
@@ -1,6 +1,6 @@
 import jsonpath from "../util/jsonpath";
 
-import { selectFragment } from "./formData";
+import { selectFragment } from "./formData"; // eslint-disable-line
```
heymatthenry, line 3: Just pointing this out here but it applies in a few other spots: it might be helpful to specify which rule is being disabled for a particular line, just for the sake of being intentional about which linting errors can be safely ignored and which actually need to be fixed.

andrewadcock: Good thinking. I've updated to include what we are skipping.

heymatthenry (approved):
LGTM

Issue #693, opened 2020-09-29
Section 0 - Help Desk Link (bug)


cjhskippy:
The help desk link in section zero is appearing as code rather than a clickable link:

<img width="590" alt="Screen Shot 2020-09-28 at 5 23 54 PM" src="https://user-images.githubusercontent.com/1299543/94498671-94115300-01af-11eb-8496-8797fb6ad4cd.png">

tadhg-ohiggins:
We don't really want to allow full HTML rendering of whatever is in the JSON, so I think for this we need a `fieldset_type` of `link`, where the value of `link` is an array of dictionaries with `text` and `url` properties. The frontend would go through the array and find (in the `label` value) the text indicated by the `text` value, then replace it with an `a` element with the appropriate `href`.
Others might have different ideas for this, though.

mgwalker:
Ooh, doing inline (rather than block) fieldsets would be an entirely new thing. There's some complexity there... But I do 100% agree we should not render HTML from the JSON.

My question would be how many places does this occur? If this is the only spot, I might say special case it in the code because that could be faster than changing the JSON. (I.e., if you're in section 0, we manually parse the text to pull out the email address and turn it into a link). If there are more occurrences, though, then it's probably better to update the JSON schema.

(I'm curious why the value would be an array of links instead of just one?)

PR #781, opened 2020-10-07
Support markdown-style links

mgwalker:
Since we already have a `<Text>` layout component, it made sense to me that the easiest way to add links to section 0 was to parse them out of the text inside that `<Text>` component. Rather than try to parse HTML-style links, which is messy at best, I opted to switch to Markdown-style links. The upshot is that any part of the app that uses the `<Text>` layout component can now have links embedded by using `[link text](link URL)` format.

- addresses #693 
   
   screenshot of link in section 0

### DIFF
--- a/frontend/react/src/components/layout/Text.js
+++ b/frontend/react/src/components/layout/Text.js
@@ -1,14 +1,52 @@
 import React, { useMemo } from "react";
 
+const linkRegex = /\[([^\]]+)\]\(([^)]+)\)/g;
+
+const parseLinks = (str) => {
+  let links = linkRegex.exec(str);
+  if (links) {
+    // Copy off the string locally so we can modify it. We'll be chopping bits
+    // off as we go along.
+    let remainingStr = str;
+    const parts = [];
+
+    while (links) {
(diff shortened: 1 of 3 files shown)


frontend/react/src/components/layout/Text.js
```diff
@@ -1,14 +1,52 @@
 import React, { useMemo } from "react";
 
+const linkRegex = /\[([^\]]+)\]\(([^)]+)\)/g;
```
heymatthenry, line 3: Just curious about why this is outside of the parsing func since. Were you thinking this might be used to detect links outside of the parser but then ended up just parsing all the text?

andrewadcock (approved):
Markdown is a smart choice!

heymatthenry (approved):
LGTM!

frontend/react/src/components/layout/Text.js
mgwalker: Mainly I put it out here so it didn't get recreated over and over. It's a habit leftover from the days when creating things like regexes was expensive. I don't know if that's true anymore...

PR #789, opened 2020-10-07
Tadhg/merge 765 branch

tadhg-ohiggins:
Now with user roles!
Admins can switch between states
With the URL.

- Adds Okta authentication to the backend endpoints.
- Derives user information from what we get from Okta.
- Creates users in the DB based on that information.
- New views that allow admin users to switch between the states we have data for (e.g. `/views/sections/MA/2020/0`).
- Users can only access backend endpoints that they're associated with.
- Table of contents works for both section views.
- Caching for some Okta requests.
- Okta URLs defined in environment variables.
- Various other tweaks/fixes associated.

Primarily addresses #206, with some other fixes to come in #794


frontend/react/src/actions/initial.js
```diff
@@ -20,31 +21,44 @@ export const getAllStatesData = () => {
...
+      dispatch({ type: LOAD_SECTIONS, data: [] });
+      throw err;
+    });
 
     dispatch({ type: LOAD_SECTIONS, data });
   };
 };
 
-export const loadUserThenSections = ({ userData }) => {
+export const loadUserThenSections = ({ userData, stateCode }) => {
```
mgwalker, line 51: Walking backwards through the code, it looks like this `stateCode` is always, forever going to be undefined? If that's correct, can we just yoink it out?

mgwalker: Oh, never mind. I see it's used in `InvokeSection`. I was looking for `<SecureInitialDataLoad>` rather than just a function execution, but 'tis what 'tis.

frontend/api_postgres/carts/settings.py
```diff
@@ -70,12 +70,26 @@
 
 REST_FRAMEWORK = {
     'DEFAULT_AUTHENTICATION_CLASSES': [
-        'rest_framework.authentication.SessionAuthentication',
+        'carts.auth_dev.JwtDevAuthentication',
```
timoballard, line 73: we may want to create an ticket as a reminder that this needs to be changed to `carts.auth.JwtAuthentication` in the productions settings

frontend/react/src/store/stateUser.js
```diff
@@ -1,67 +1,69 @@
...
-      return state;
-  }
-}
+// ACTION TYPES
+const STATE_INFO = "STATE_INFO";
+const USER_INFO = "USER_INFO";
+const PROGRAM_INFO = "PROGRAM_INFO";
+
+// ACTION CREATORS
+export const getUserData = (userObject) => {
```
mgwalker, line 7: Should these action creators be `setUserData` etc?

frontend/api_postgres/carts/carts_api/model_utils.py
```diff
@@ -66,7 +70,97 @@
...
+    ("admin_user", "Admin User"),
     ("co_user", "Central Office User"),
+    ("state_user", "State User"),
+    ("temp_user", "Temporary User"),
+)
+
+JOB_CODES_TO_ROLES = {
+    "IDM_OKTA_TEST": "state_user",
+    "CARTS_Group_Dev": "admin_user",
+}
```
mgwalker, line 83: Eventually it'd be good to pull this out to a config or put in the database or something so it's not hard coded.

frontend/api_postgres/carts/carts_api/model_utils.py
```diff
@@ -66,7 +70,97 @@
...
+    def shape(val: list) -> dict:
+        with_ldap_keys = dict(group_dc(val))
+        return {
+            "job_code": with_ldap_keys["cn"],
+            "organizational_unit": with_ldap_keys["ou"],
+            "domain_component": with_ldap_keys["dc"],
+        }
+
+    delimited = entry.replace(",cn=", ";SPLIT;cn=")  # crude, but ¯\_(ツ)_/¯
+    raw_entries = delimited.split(";SPLIT;")
```
mgwalker, line 131: This is to preserve the `,cn=`, I reckon? Because you split on that directly, it would be consumed?

frontend/api_postgres/carts/oidc.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def metadata(key):
+    discovery_doc = fetch_discovery_doc()
+    return discovery_doc[key]
+
+
+def verify_token(token, pub_key):
+    jwk = JWK.from_json(json.dumps(pub_key))
+    return jwt.verify_jwt(token, jwk, ['RS256'], checks_optional=True)
```
mgwalker, line 56: This whole file is fancy. I like how it works!

frontend/api_postgres/carts/permissions.py
```diff
@@ -0,0 +1,27 @@
+from rest_framework import permissions
+
+
+class StateViewSectionPermission(permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        user, view_name = request.user, view.get_view_name()
+        if request.user.has_perm("carts_api.view_section"):
+            print(f"User {user} has admin permissions for {view_name}.")
```
mgwalker, line 8: Is this `print` accurate about admin permissions?

frontend/api_postgres/carts/settings.py
mgwalker: Yes please. We should also make sure our devops friends have a heads-up about it.

mgwalker (approved):
I see there are still changes happening here, but what's here now looks good to me. A couple of potential cleanup questions ("admin" permission? "setUser"?) Also please update the PR description with the list of issue #s that are covered here, since there are a few. Otherwise, 🚀

frontend/api_postgres/carts/permissions.py
tadhg-ohiggins: Kind of, in that only admin and co_user users should have permission for the entire view, but changing it to make it more clear.

PR #794, opened 2020-10-08
More fake users, ability to associate usernames with states, status endpoint

tadhg-ohiggins:
Add more fake users,
Add state association,
And add state status.

- Ensure that all info about a user is updated from the EUA info upon login.
- Add `dev-co_user` and `dev_admin` fake users.
- Add model and endpoint for associating usernames with states; make sure that association takes place on login; default to MA for now for users who aren't admins/co_users and who don't have states in the table yet.
- Add model and endpoint for state status, with provisional list of statuses.

Builds on #789 and shouldn't be reviewed before that's merged.


tadhg-ohiggins:
Just converting this to draft while the ongoing issues on mdctcartsdev.cms.gov are worked out; not sure what's going on there and don't want this getting merged in the middle of things.

timoballard (approved):

PR #803, opened 2020-10-08
Remove autologout on auth fail

tadhg-ohiggins:
While automatic,
It doesn't always make sense
So we remove it.

Removes logging out as soon as auth fails, because it makes troubleshooting hard.


mgwalker (approved):
temp fix

PR #807, opened 2020-10-09
scale back api in prod to 1 task

mdial89f:
We want to test a failure mode.
Scaling back the api service from 3 to 1 in prod environment.


mdial89f:
doug figured out why this all was failing.  we no longer need to test this failure scenario.  closing

Issue #751, opened 2020-10-03
Remove "Share" button from form sections


cjhskippy:
All form sections currently include a "Share" button at the bottom of the page, near the "Next" and "print" buttons.  This share button doesn't currently work (and we don't expect to be able to make it work before launch).  Let's remove it from these pages so that it doesn't confuse users.

You can find an example at the bottom of this page: https://mdctcartsdev.cms.gov/sections/2020/0?dev=dev-ak

<img width="261" alt="Screen Shot 2020-10-03 at 10 13 01 AM" src="https://user-images.githubusercontent.com/1299543/94997621-42771880-0561-11eb-8ae3-6b47fbc8992d.png">

AC:
- [ ] Section pages no longer include a "Share" button.

PR #811, opened 2020-10-09
Remove share button

mgwalker:
- removes the share button from all the form pages
- also removes lingering share logic, since we don't need it now
- addresses #751


heymatthenry (approved):
Hooray for ripping out code!

PR #820, opened 2020-10-14
Update SEDS Query

jgillis1:
Translate SQL to postgres. Save a copy of it here for posterity.


jgillis1:
Out of date at this point.

PR #846, opened 2020-10-18
Improve performance by being smarter about paths

mgwalker:
I was playing around with [Redux Toolkit](https://redux-toolkit.js.org/) this weekend for funsies, and I was using this project as my playground. It was still slow to load bigger sections (especially 3C) and I had a sudden realization: given an ID, we can short-circuit a _*lot*_ of the jsonpath lookup.

IDs look like this:
```
year-section-subsection-part-question...
```

By making sure sections are stored in ordinal order, we can then build a path with this information, like so:

```
?..formData[section].contents.section.subsections[subsectionNumber].parts[part - 1]
```

(`subsectionNumber` is computed by converting the subsection letter to its ASCII value and subtracting 97, the ASCII value for `a`. This is a naive approach and doesn't handle the possible case where a subsection has two letters. But that doesn't exist for now, and I guess the answer would be to convert each letter to its ASCII value separately, multiply the first by 26, add the two together, and subtract 97.)

So rather than having to check IDs all over the place, we're instructing the jsonpath query thingy which exact array indices to look at, and we end up having to look at a whole lot less stuff.

That's what this PR does. Try it out! It seems about 10x faster in Chrome in my experience.


frontend/react/src/store/formData.js
```diff
@@ -4,12 +4,25 @@ import { SET_FRAGMENT } from "../actions/repeatables";
...
+  return 0;
+};
+
 const initialState = [];
 
 export default (state = initialState, action) => {
   switch (action.type) {
     case LOAD_SECTIONS:
-      return action.data;
+      return action.data.sort(sortByOrdinal);
```
mgwalker, line 25: Sort here. Then we don't have to do it elsewhere, and we can make assumptions about mapping section numbers to indices.

frontend/react/src/util/jsonpath.js
```diff
@@ -15,6 +15,47 @@
...
+    // Now see what bits we have and assemble the pieces.
+    const pathParts = ["$."];
+    if (section) {
+      // Sections start at 0, so we don't have to do anything special with them.
+      pathParts.push(`.formData[${+section}].contents.section`);
+    }
+    if (subsection) {
+      // Subsections are lowercase letters, starting at a. Subtract the ASCII
+      // value for lowercase a (97) to compute the index.
+      const subsectionNumber = subsection.toLowerCase().charCodeAt(0) - 97;
```
tadhg-ohiggins, line 39: Am I missing a reason why we can't just use the index of the letter in `letterMarkers` instead of doing the ASCII math?
https://github.com/18F/cms-carts-seds/blob/master/frontend/react/src/store/formData.js#L105

tadhg-ohiggins:
This looks like a great idea. I'll check its speed and look more closely at it tomorrow; for now I just have the one comment I made.

mblevins-collabralink:
I was actually wondering over the weekend if we could use something like this strategy to speed up the page loads.

Great minds, I guess - good find, @mgwalker

frontend/react/src/util/jsonpath.js
mgwalker: Good call. I pulled that list into a separate file so it'd be easier to share (otherwise we'd end up with a cyclical dependency) and switched it up. 👍

tadhg-ohiggins (approved):

Issue #300, opened 2020-08-05
Login Page


TaraWilliamsPGH:
As a CARTS user I need to be able to log into the application so I can do my job.

AC:

- [x] Login accepts user credentials (handled by OKTA; we need to manage credentials on the DB side to understand level of permissions)

- [x] Validates user permissions (type of user pulled from DB)

- [x] Directs user to appropriate homepage landing (meaning state user, central office, invalid user, etc)

cjhskippy:
Taking notes here about some other important considerations, even if we decide to put them in one or more separate stories:

- [ ] What's the UX for failed logins?
- [ ] Where a user can get help if their login isn't working

TaraWilliamsPGH:
NOTE: Laura/Seth have already done some work on this...please verify first.

mgwalker:
@tadhg-ohiggins Are you currently working on the UI part of this? If not, I can pick it up.

tadhg-ohiggins:
@mgwalker Nope, all yours

PR #849, opened 2020-10-19
Add separate landing pages for different user roles

mgwalker:
Another step towards completing #300.

- adds distinct landing pages for state users (same as current default landing page), admin users (screenshot below), and business/CO users (these two currently get the same page)
- tweaks routing so that only admin users get admin routes, state users get state routes, etc.

Some things we'll need before #300 can be closed:

- an API endpoint for business/CO users to fetch a list of all reports
- the `CMSHomepage` component is updated to show that list of reports
- `SecureInitialDataLoad` will need to be smarter to handle user types
- we'll want to delay loading whole reports until the bus/CO user clicks it, for performance reasons

@StephanieCoppel Should business/CO users only see certified reports, or should they see all reports along with their statuses? Same question for whether or not they can review all reports, or just certified ones.

---

Admin page:
screenshot of admin landing page

### DIFF
--- a/frontend/react/src/store/stateUser.js
+++ b/frontend/react/src/store/stateUser.js
@@ -33,9 +33,9 @@ const initialState = {
   imageURI: `${process.env.PUBLIC_URL}/img/states/ny.svg`,
   formName: "CARTS FY",
   currentUser: {
-    role: "admin",
-    state: { id: "NY", name: "New York" },
-    username: "karen.dalton@state.gov",
+    role: false,
+    state: { id: "", name: "" },
+    username: "",
   },
 };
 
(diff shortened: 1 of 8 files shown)


frontend/react/src/store/stateUser.js
```diff
@@ -33,9 +33,9 @@ const initialState = {
   imageURI: `${process.env.PUBLIC_URL}/img/states/ny.svg`,
   formName: "CARTS FY",
   currentUser: {
-    role: "admin",
-    state: { id: "NY", name: "New York" },
-    username: "karen.dalton@state.gov",
+    role: false,
+    state: { id: "", name: "" },
+    username: "",
```
mgwalker, line 38: Clear out the default user, so we don't get that flash of `karen.dalton` anymore. Also prevents us from showing admin/state/CMS routes to people who don't have permissions, even briefly.

heymatthenry (approved):
LGTM!

PR #850, opened 2020-10-19
Add a page for CMS CO and business owner users

mgwalker:
All CMS users get the same page for now: a listing of states that have a form with a status, and links to review them.

screenshot of CMS user page

Also:
- rejiggers the initial data loading so that users and sections can be loaded completely separately
   - user info is required in order to load the higher-order routing components that then load sections, which guarantees the right order of operations
- updates the `<SecureInitialDataLoad>` so that it calls Okta login itself if the user is not logged in and auth is not pending. Unclear why this situation ever happens, but when it does, it becomes impossible to get an automatic login to "just work." So instead of relying on it "just working", "just manually do it."
- widen the form table so that state names will fit in the first column for CMS users


tadhg-ohiggins:
Very minor issue: from what I can tell, the official name for the status after `not started` is `in progress`, and there is no `started` status.

tadhg-ohiggins (approved):

PR #852, opened 2020-10-20
Use routing links instead of anchor tags

mgwalker:
For the dashboards, use `<Link>` components instead of `<a>` tags so that routing happens within react-router instead of causing a page reload. Fewer trips to the server, should keep things moving along.


heymatthenry (approved):
Good thinking!

PR #854, opened 2020-10-20
Add spinner while waiting for form pages to load

andrewadcock:
Created a spinner component and styled. 

The component is added to WrapSecurity. This is initiated via dispatch in `intial.js`, shown only on `/sections/` pages, and is terminated in `Part.js`.


cl-adavyd (approved):

mblevins-collabralink (approved):

frontend/react/src/components/layout/Part.js
```diff
@@ -91,7 +91,13 @@ const mapStateToProps = (state, { partId }) => {
...
   };
 };
 
-export default connect(mapStateToProps)(Part);
+const mapDispatchToProps = (dispatch) => {
+  // End spinner, started in initial.js
+  dispatch({ type: "CONTENT_FETCHING_FINISHED" });
+};
+
+export default connect(mapStateToProps, mapDispatchToProps)(Part);
```
mgwalker, line 103: I think using `mapDispatchToProps` for side-effects would be an anti-pattern. A future dev wouldn't know to look here for anything other than mapping actions to component properties. I get why you went this route, but there's another way! 😄

frontend/react/src/actions/initial.js
```diff
@@ -129,6 +129,9 @@ export const loadForm = (state) => async (dispatch, getState) => {
   const { stateUser } = getState();
   const stateCode = state ?? stateUser.currentUser.state.id;
 
+  // Start isFetching for spinner; turned off in Part.js
+  dispatch({ type: "CONTENT_FETCHING_STARTED" });
+
   dispatch(loadSections({ userData: stateUser, stateCode }));
   dispatch(getStateStatus({ stateCode }));
   dispatch(getAllStatesData());
```
mgwalker, line 137: All of these dispatches return promises, so you can do something like:

```js
dispatch({ type: "CONTENT_FETCHING_STARTED" });

await Promise.all([
  dispatch(loadSections({ userData: stateUser, stateCode })),
  dispatch(getStateStatus({ stateCode })),
  dispatch(getAllStatesData()),
]);

dispatch({ type: "CONTENT_FETCHING_FINISHED" });
```

mgwalker (commented):
TIL an animated gif will keep animating even when the UI thread is blocked. Nice!

frontend/react/src/actions/initial.js
andrewadcock: Thank you! That's much cleaner and allowed me to remove the code from part.js and also the check for what pages show the spinner. Thanks!

mgwalker (approved):

PR #859, opened 2020-10-20
Bug/computed fields blank

TGriesemer:


frontend/react/src/util/synthesize.js
```diff
@@ -155,11 +154,22 @@ const formula = (targets, providedFormula, precision) => {
...
+  values.map((value) => {
+    if (value !== null && value !== "") {
+      allNull = false;
+    }
+    return "";
+  });
+  return allNull === false
+    ? values.reduce((acc, value) => acc + +value, 0)
+    : "";
+};
```
mgwalker, line 172: Rather than using `Array.map` for side-effects, you can use `Array.some` to verify that at least some of the values are valid:

```js
const hasNumbers = values.some(value => value !== null && value !== "");
if (hasNumbers) {
  return values.reduce((acc, value) => acc + +value, 0);
}
return "";
```

TGriesemer: Thanks for the suggestion!

mblevins-collabralink (approved):

PR #862, opened 2020-10-21
Bugfix/ Daterange 

### LINKED ISSUE #508: Bug: DateRange Validation
Check that validation is able to bypass when date groups are empty and allow for individual pairs to submit. 

Also, when entering a date, if you put in too many digits into the year, it is not possible to correct the error.
Reproduction steps:
1. Go to https://mdctcartsdev.cms.gov/sections/2020/02?dev=dev-ak
2. Answer "yes" to part 2, question 3.
3. For question 3b start date, enter 02 / 2020
4. For question 3b end date, enter 02 / 20200 and hit tab to change focus to the next question
5. Navigate back to the year field of the end date. Attempt to delete one of the characters. It will remain unchanged.

AC: 
- [ ]  If one date group is empty ( both fields): return empty string for that grouping
- [ ] If both date groups have been removed; remove error messaging
- [ ] If both date group have been removed; send null to redux
- [ ] Only allow max of 2 characters for the Month 
- [ ] Only allow max of 4 characters for the Year

alewoo28:
**Overview:**
Fixed DateRange component bugs. The first bug wouldnt allow changes to the input after all 4 fields were validated. Several other bugs cropped up after fixing the first one. Issue #508 

**Files changed (1):**
`frontend/react/package-lock.json` / `frontend/react/package.json`
Added [React UUID](url) 
The errors that appear above the input fields are from an array of strings being mapped through in state. Using the element's index in the array as the key for each element was maybe responsible for wonky bugs as that array was constantly being added to and removed from. This package creates a UUID for each  element. This should not be any significant performance hit as the array will never have a length greater than 4. 

`frontend/react/src/components/layout/DateRange.js`
The validateMonth isnt in the body of the class as a method because it never utilizes **_'this'_**.
Data from input to redux ( for the start range) goes  
onChange (handleInput) > onBlur (validateStartInput) > validateMonth() & validateYear() > checkChronology() [data fully validated] > props.onChange(), sent to redux. 

This is all in one commit because I had to create a new branch and copy over the contents of DateRange.js. I made a commit that edited a utility function file and with the new rules on the linter it created a chain reaction where I would have had to edit (or delete) 2 other files (which interacted with the utility file) in order to satisfy the linter and making changes to those files introduced the risk of having to change additional files that they interact with, none of which have anything to do with this ticket. 

To test the DateRange go to Section 2, question 3. Answer 'yes', and 3b has a DateRange


cl-adavyd (approved):

frontend/react/src/components/layout/DateRange.js
```diff
@@ -1,296 +1,332 @@
...
+      const startDate = storedValue[0].split("-"); // ie: ['2019', '11', '01']
+
+      // Returns an array with the values of the start date, split
+      const endDate = storedValue[1].split("-"); // ie: ['2020', '09', '01']
+
+      const monthStartValue = startDate[1]; // '11'
+      const yearStartValue = startDate[0]; // '2019'
+
+      const monthEndValue = endDate[1]; // '09'
+      const yearEndValue = endDate[0]; // '2020'
```
mgwalker, line 72: If you wanedt, this could be shortened with array destructuring:

```js
const [ startYear, startMonth ] = storedValue[0].split("-");
const [ endYear, endMonth ] = storedValue[1].split("-");
```

frontend/react/src/components/layout/DateRange.js
```diff
@@ -1,296 +1,332 @@
...
+
+      this.setState({
+        monthStart: monthStartValue || "",
+        monthEnd: monthEndValue || "",
+        yearStart: yearStartValue || "",
+        yearEnd: yearEndValue || "",
```
mgwalker, line 78: Someone on the CL team started using null coalescing elsewhere, so it'd make sense to either use it everywhere, or take it out of those other places. I think it's cleaner and implies the intention better, but I don't feel very strongly about it, to be honest. Just flagging it for the team to think about. 😄 

```suggestion
        monthStart: monthStartValue ?? "",
        monthEnd: monthEndValue ?? "",
        yearStart: yearStartValue ?? "",
        yearEnd: yearEndValue ?? "",
```

frontend/react/src/components/layout/DateRange.js
```diff
@@ -1,296 +1,332 @@
...
+
+    // Ensure that all 4 fields are filled in
+    if (monthStart && monthEnd && yearStart && yearEnd) {
+      // Turn the input into date objects for easy comparison
+
+      monthStart = monthStart.padStart(2, "0");
+      monthEnd = monthEnd.padStart(2, "0");
+
+      const startDate = new Date(yearStart, monthStart - 1);
+      const endDate = new Date(yearEnd, monthEnd - 1);
```
mgwalker, line 112: I'd suggest moving this to after you create the `Date` objects. I'm like 98% sure it's fine as you have it, but it feels weird to me to pad the number with zeros and then do math on it. It ***feels*** like a bug (even though I'm pretty sure it's not), which might cause a future version of you to slow down when looking at this code.

```suggestion
      const startDate = new Date(yearStart, monthStart - 1);
      const endDate = new Date(yearEnd, monthEnd - 1);

      monthStart = monthStart.padStart(2, "0");
      monthEnd = monthEnd.padStart(2, "0");
```

alewoo28: Makes sense, taken care of!

frontend/react/src/components/layout/DateRange.js
alewoo28: Done!

Issue #450, opened 2020-08-27
Document status - In Progress


StephanieCoppel:
CARTS Reports can have the following statuses: 
**In Progress** - report is available and open to edits by states, can be viewed by Business Owners and CHIP Project Officers

ACs:

- [x] documents can exist in progress

- [x] states can edit report

- [x] CHIP Project Officers and Business Owners can view but not edit

- [x] In this status, the state user has the option to 'certify and submit' to move the document on to the next status, Certified and Submitted 

- [x] this status can also exist when the CHIP Project Officer 'uncertifies' a document, it moves back to this in progress status

**WORKFLOW MURAL:** https://app.mural.co/t/upmc4555/m/upmc4555/1601924190266/917889a4d9489374a4f5fd0a043f9ed41f27bb8c

cjhskippy:
Two issues observed during demo:
- Verify that frontend will update the status to "In Progress" when a user makes a change to an "Uncertified" issue
- co_user was able to make a change to a document that was "In Progress"

PR #864, opened 2020-10-21
Prevent any users other than state users from saving forms

tadhg-ohiggins:
You a state user?
No? Then don't try to edit.
K thx bye seeya

Changes to the contents of the form can only be made by users with the `state_user` role.
Related to the third item in #450


mgwalker (approved):
I guess this makes sense if we're embracing the tyranny of authorized access. 😂

PR #866, opened 2020-10-21
Make new status entry when changing to in_progress from another status

tadhg-ohiggins:
If status changes
We should capture who did it
So they cannot hide

Creates a new entry for a status change if we're actually changing the status value, rather than modifying an existing entry. If the change is just an edit, with no status change, we update the last_changed timestamp and make no other changes to the status entry (and don't make a new status entry).


mgwalker (approved):
high five

Issue #841, opened 2020-10-16
Frontend - editing permissions tied to status of the form


StephanieCoppel:
To capture the frontend portion of the ability to edit the form ONLY in 'in progress' status. while the form is in any other status, the form should not be editable.

ACs:

- [x] editing of the form is possible only in the 'in progress' status

- [x] all other form statuses, the form should not be editable and response boxes should be grayed out to illustrate it's not available to edit

PR #867, opened 2020-10-21
Disable questions in the form if it isn't in an editable status

mgwalker:
- make questions `disabled` if the form is not editable
- disable repeatables add/remove buttons if the containing question is disabled
- update the landing page so that it shows the actual form status and either an "Edit" or "View" link, depending on the status
- addresses #841


heymatthenry (approved):
LGTM!

Issue #764, opened 2020-10-05
fix alert on Section 3I


erinzstrenio:
Alert under Section 3I, Part 2, question 2 should only display if users selected "no" for this question. 

<img width="723" alt="Screen Shot 2020-10-05 at 2 33 12 PM" src="https://user-images.githubusercontent.com/46569779/95123552-c2df7a00-0717-11eb-9e1a-c4d1d263d769.png">

mgwalker:
This seems to be fixed? https://mdctcartsdev.cms.gov/sections/2020/03/i?dev=dev-ma

erinzstrenio:
Hm, it's still showing up for me before the question is answered or if "yes" is selected. It should only appear if "no" is selected.

mgwalker:
Weird! Alrighty then! Let's keep this here until we can confirm that it's fixed, then.

StephanieCoppel:
it is not fixed for me either... lets keep this :)

PR #870, opened 2020-10-21
Fix logic for (non)display of §3(i) alert message

heymatthenry:
This PR just makes a small change to the show/hide logic in §3(i) to correctly hide the HSI questions when user indicated they don't have a HSI program.

Addresses #764


heymatthenry:
@mgwalker I'll make a new story for the other changes we discussed. Just wanted to get this change into the review queue.

mgwalker (approved):
that was too easy

Issue #840, opened 2020-10-16
Disable form fields for CMS users (business owners and project officers)


cjhskippy:
As a CMS business owner or project officer, I want clear indication that the form cannot be edited so that I don't get confused when my attempts to change form fields are rejected.

AC:
- [x] When CMS users (bus_user and co_user) view any form section, the input fields are disabled.
- [x] Form editability for state users should remain unchanged:
  - [x] When the form status is "Certified" or "Approved", all form fields are disabled, providing a visual indicator that you can't make changes.
  - [x] When the form status is "In progress," users are able to edit form fields normally, except for the ones that are always disabled in the "Basic Info" section.

PR #871, opened 2020-10-21
Disable form fields for CMS users

mgwalker:
- makes form fields disabled for CMS users
- updates the CMS dashboard so that it shows friendly statuses instead of the all-lowercase, snake-cased internal names
- addresses #840


heymatthenry (approved):
👍 🥇

PR #882, opened 2020-10-23
Re-enable authentication

### LINKED ISSUE #480: Re-enable authentication

timoballard:
This splits the Django settings into separate files for dev and prod, where the prod settings disable the authentication pass through. The settings file to be used in each environment is controlled via an environment variable, `DJANGO_SETTINGS_MODULE`

resolves #480


tadhg-ohiggins:
Converted to draft while doing some troubleshooting, just didn't want this to get merged while in the middle of doing that!

tadhg-ohiggins (approved):

tadhg-ohiggins:
I've held off on merging it because we haven't gotten confirmation that the pen tester accounts are working, or that we definitely have someone whose EAU ID is associated with an app admin role (that is, someone who can set other user roles when necessary). Once those two things are in place, this can be merged.

mblevins-collabralink (approved):

Issue #884, opened 2020-10-27
Section Titles Not Showing/Showing Twice


andrewadcock:
Occasionally, Section titles are being printed twice or not at all. ![duplicate-titles.png](https://images.zenhubusercontent.com/5f280c57a9416e9310a2bc6c/d6c3ac0e-aae2-44cb-8db2-41468f68f78a)

This may be affected by user information; using the query parameter (?dev=dev-ak) seems to change the result.

AC:

- [ ] Section Titles should show exactly ONCE per page

Issue #872, opened 2020-10-22
Section 4 title shows up twice 


erinzstrenio:
"State Plan Goals and Objectives" shows up twice at the top of Section 4

![Screen Shot 2020-10-22 at 10 05 24](https://user-images.githubusercontent.com/46569779/96891654-7e114e00-144e-11eb-8fd6-864eeed4a2f8.png)

PR #892, opened 2020-10-28
Fix title duplication

jgillis1:
Related to #872 and #884
-Section 2
-Section 4

### DIFF
--- a/frontend/api_postgres/utils/section-schemas/backend-json-section-2.json
+++ b/frontend/api_postgres/utils/section-schemas/backend-json-section-2.json
@@ -6,9 +6,9 @@
     "valid": null,
     "ordinal": 2,
     "type": "section",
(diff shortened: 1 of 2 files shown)


andrewadcock:
Looks like the title doesn't show when using the state query parameter BUT it shows fine without. The section 2 title in the navigation now says undefined. I'm not sure if these are directly related to your changes or another bug.

titles

jgillis1:
When using the state hack (?dev=dev-xx), the title of the Section you initially navigate to does NOT display. If you navigate to other Sections, you will see their titles however the title of the Section you landed on appears to never display. As this is an issue using the hack, I do not believe this is a real bug that needs to be addressed. It's merely a byproduct of the work-around to get developers access in their local environments without needing an account set up.

andrewadcock (approved):
LGTM!

PR #898, opened 2020-10-30
Print Functionality for Logged in Users

andrewadcock:
Updated code to accurately pull logged-in user data. Previously print was pulling initialState user information and not updating with currentUser data.

An odd change: useEffect was throwing errors with dispatch returning a promise. The solution was to move the data loading functionality into a function that is async and have use effect call it.

Close print dialogue box when:

- Print `This Section` is selected
- Print `Entire Form` is selected
- User clicks anywhere else on the page (that's not the print dialogue box)


jgillis1 (approved):
Looks good!

PR #910, opened 2020-11-03
Feature/add help button

### LINKED ISSUE #265: Add help button to top navigation in dropdown menu
As **state or federal user**, I need to **where to get help with using CARTS** so that I can **get support with errors or questions quickly.**

As per the last round of usability testing with state users, a couple of users were unable to find a way to get help with CARTS and didn't notice the email address in the footer. State users who were unable to find a help button or link expected it to be somewhere in the header. In the last round of usability testing with federal users, help button was added to the top navigation dropdown and all users were able to find it. 

- [ ] Add a help button to the dropdown menu in the header.
- [ ] Keep the help email visible in the footer 

image

alewoo28:
**Overview:**
Added help desk button to top navigation in dropdown menu. This is the same link present in the footer, it opens up an email window with cartshelp@cms.hhs.gov as the recipient. Also fixed formatting on the logut button. Closes #265 

**Files changed (3):**
**frontend/react/src/components/layout/Header.js**
Changed display name above dropdown to user's email address instead of EUA ID, as per design doc. Extracted email from redux props. Removed unused attribute. 

**frontend/react/src/components/layout/Logout.js**
Added logout icon directly to logout component

**frontend/react/src/scss/_header.scss**
Added various styling to UL elements, logout, helpdesk and manage-account. Added helpdesk icon. 

The design doc asks for email addresses to  be displayed above the top nav menu, but I wonder if we should instead stick to displaying EUA IDs as my name alone is very long "alexis.woodbury@cms.hhs.gov" whereas EUA IDs would be a set length.


jgillis1:
@StephanieCoppel Question for you before we merge this. #265 shows Email instead of EUA ID at the top for the logged in user. Which would you prefer to see?

jgillis1 (approved):
Getting Stephanie to weigh in on email vs. EUA ID but otherwise, the code looks good.

jgillis1:
Stephanie replied that email is preferred so this is good to merge.

Issue #787, opened 2020-10-07
AREMAC Endpoints


TaraWilliamsPGH:
As a CARTS DSCP user I need to be able to do reporting within AREMAC so I can complete my responsibilities. 

AC:

- [ ] Create an API to allow for data access into CARTS from AREMAC

- [ ] Create an endpoint to allow how CARTS / AREMAC will access the API

Tech Note:

- [ ] Can we use existing endpoints for the states?

andrewadcock:
AREMAC has requested access to pull directly from S3
They would also like parquet instead of JSON


From AREMAC
CARTS Data Ingestion - Requirements
CARTS team to AREMAC

**How will AREMAC connect?**
AREMAC team uses s3 access points. Event-based. SQS aka s3 put notification. 
**Does AREMAC have a preferred or required data format?**
Parquet is preferred
**How will CARTS authenticate requests from AREMAC?**
VPC pairing via access points

PR #914, opened 2020-11-03
[WIP] Add json endpoints for state report

andrewadcock:
Satisfies: #787 

The result of this PR is JSON support for the report endpoints. If the correct header is present return JSON else return html.

The biggest item here is adding @csrf_exempt to these method. There is no need to authenticate a user when pulling this data as no PII is included. I'm not sure what other implications this might have.

You'll note that I'm pulling the first result from an array in a couple places, that's to ensure that the latest item is displayed. State Status does not update records, instead it adds new with a timestamp.

I have some minor concerns about the resulting data structure so please review the output as well.

Endpoints:
**Single State**
/report/states/int:year/str:stateAbbrev/
Returns all form data. Questions with Entries, status, program type

**All States**
/report/states/int:year/
Returns all states (alpha) with Questions with Entries State Name, State Abbrev, program type

**All State Status**
/report/states/status/int:year/
Returns all states (alpha) status for given year. Includes: state abbrev, year, status, last_changed


jgillis1:
Closing for now as the plan is for AREMAC to get the CARTS data via BigMac

PR #933, opened 2020-11-11
Bugfix: Numerical logic 

alewoo28:
**Overview:**
Create validation functionality for the Range component. Issue[ #2383](https://qmacbis.atlassian.net/browse/OY2-2383)
Ranges with Percentage and Money fields needed to be checked so that the "start" values were less than the "end" values. 

**Files changed (3):**
`frontend/react/src/components/fields/Integer.js`
Commas in the value that come from the field mask or user input are stripped. The value is parsed and confirmed to be a number. The value is sent back to the component as a string. 

`frontend/react/src/components/fields/Ranges.js`
State is used to keep track of range values and errors. 
Added some useful comments to existing code. 
validateInequality function written to validate two fields against one another if they are both present in the values array and neither of them is null. 
changeStart & changeEnd update local state before sending their values back up to the Ranges component. 
Input value is now being read from local state instead of straight from redux. 

`frontend/react/.eslintrc.json`
Added the correct React version to the linter settings
Added ES6 to the linter settings

Data is being sent to redux on each keystroke, so validating one field against another before they reach redux is not a viable option without a more significant rework. This means that if a user puts in values that fail validation, they persist. The way its being handled now is that an error message is displayed on the UI and the function that generates that error will also be triggered when a user navigates back to the page.


alewoo28:
Updated the linter settings

jgillis1 (approved):

PR #934, opened 2020-11-12
Dev waf logs

ayoneeyee:
Added Terraform code to enable logging on WAF

### DIFF
--- a/frontend/aws/ecs_task_api_postgres.tf
+++ b/frontend/aws/ecs_task_api_postgres.tf
@@ -349,3 +349,127 @@ resource "aws_wafv2_web_acl_association" "apipostgreswafalb" {
   resource_arn = aws_alb.api_postgres.id
   web_acl_arn  = aws_wafv2_web_acl.apiwaf.arn
 }
+
+# # ============== Data ====================
+data "aws_caller_identity" "current" {}
+
+locals {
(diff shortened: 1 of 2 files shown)


mdial89f (commented):
You might consider setting the buckets for val/prod in code, in a local variable table.
see https://github.com/CMSgov/cms-carts-seds/blob/master/frontend/aws/ecs_task_api_postgres.tf#L3-L5
and https://github.com/CMSgov/cms-carts-seds/blob/master/frontend/aws/ecs_task_api_postgres.tf#L50

So you could have 
`
locals {
  waf_logging_bucket = {
    prod: "<whatever the prod bucket name is>",
    val: "<whatever the val bucket name is>",
    master: "cms-cloud-730373213083-us-east-1-legacy"
  }
}
`

and then later 
`lookup(local.waf_logging_bucket, terraform.workspace, local.waf_logging_bucket['master'])
`

It let's us use the dev/master bucket by default for non-explicitly-named branches, and it keeps us from having to override the bucket variable explicitly when deploy val/prod

mdial89f (commented):
I left a comment in conversation, curious to hear your thoughts on it

ayoneeyee:
Makes sense Mike. I know its not fancy, just wanted to get it to work first. Let me update it. 
Thanks for the tips.

PR #939, opened 2020-11-17
Remove updating is_active

jgillis1:
Based on discussions had today we do not want to update is_active to true on login


andrewadcock (approved):
Very elegant coding here ;)

PR #945, opened 2020-11-23
Feature/add state user

TGriesemer:
Created a new page AddStateUser that allows an admin to assign an eua id to a state. Created an addstateuser endpoint which uses the view AddStateUser().


jgillis1 (approved):
Please check the format of the "Add User" button on the add_state_user page. It is not the same visually as the other buttons in the app.

Other than that, this looks very clean and functions appropriately. Very straightforward. After the button is updated and conflicts resolved, this PR is good to merge.

TGriesemer:
To get this to work I had to comment out starting at line 62 in auth.py
`if _is_user_active(user_info) == False:
                print("username: ", user_info.preferred_username)
                return`
load the page and then I could uncomment it out and continue. I talked with Tony and we are going to review it on Monday.

jgillis1 (approved):

PR #958, opened 2020-12-02
Add printable textarea functionality

andrewadcock:
Satisfies: [#3284](https://qmacbis.atlassian.net/browse/OY2-3284?atlOrigin=eyJpIjoiMGJjZmEwNTJmZGJkNGY1NThkN2E0OWE5YzkyMjAzMWUiLCJwIjoiaiJ9)

There is no reasonable CSS-only solution to print textareas. Instead, I created a new div the mimics the data in the textarea. In the print styleshet, the textarea is hidden the print-helper text is shown. This works for text_multiline and text_medium.


jgillis1:
This does show all the text when the text is larger than the visual text box, however, it does not appear to word-wrap so I'm concerned that, on actually printing, things are still technically "cut off". Thoughts?

andrewadcock:
You are absolutely right. I don't think I was using the best dummy data. 

Now it's being parsed when it's added to the div properly. Line breaks, etc are in tact.

jgillis1 (approved):

PR #962, opened 2020-12-04
Default users with no role to state_user, show error if incomplete

andrewadcock:
Satisfies: [#3384](https://qmacbis.atlassian.net/secure/RapidBoard.jspa?rapidView=184&projectKey=OY2&modal=detail&selectedIssue=OY2-3384&search=3384)

Users without an entry in the `carts_api_rolefromusername` table will now default to "state_user" in app (no db change occurs). This provides the lowest amount of privileges possible.

When the user attempts to access the site without an assigned state in the `carts_api_statesfromusername` table they are now greeted with an error message containing instructions to contact the helpdesk.

### DIFF
--- a/frontend/api_postgres/carts/auth.py
+++ b/frontend/api_postgres/carts/auth.py
@@ -106,6 +106,8 @@ def _get_or_create_user(user_info):
     print(f"\n\n    $$$$$username_map is: ", username_map)
 
     if not username_map:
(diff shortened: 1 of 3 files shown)


jgillis1 (approved):

jgillis1:
As a note for testing purposes, the test user must not have a record in carts_api_rolefromusername or carts_api_statesfromusername.

PR #986, opened 2020-12-14
Filter Federal User Homepage

TGriesemer:
Addresses ticket # [2445](https://qmacbis.atlassian.net/browse/OY2-2445) for filtering the co_user/bus_user homepage. 

3 multi select drop downs were added (State, Year, Status)


frontend/react/src/components/sections/homepage/CMSHomepage.js
```diff
@@ -1,18 +1,73 @@
...
+  stateList,
+  yearList,
+}) => {
+  const statusList = [
+    { label: "Accepted", value: "accepted" },
+    { label: "Certified", value: "certified" },
+    { label: "In progress", value: "in_progress" },
+    { label: "Not started", value: "not_started" },
+    { label: "Published", value: "published" },
+    { label: "Submitted", value: "submitted" },
```
jgillis1, line 23: Do we need "Submitted"? I don't think that's a status a CARTS report can enter into.

frontend/react/src/components/sections/homepage/CMSHomepage.js
```diff
@@ -1,18 +1,73 @@
...
+    { label: "Uncertified", value: "uncertified" },
```
jgillis1, line 24: Do we need this as an option since the uncertify action puts the status back to "In Progress"?

frontend/react/src/components/sections/homepage/CMSHomepage.js
```diff
@@ -25,30 +80,132 @@ const CMSHomepage = ({ getStatuses, statuses, currentYear }) => {
...
+                  <div className="filter-drop-down-state">
+                    <MultiSelect
+                      options={stateList}
+                      value={tempStates}
+                      onChange={onSelectState}
+                      labelledBy={"State"}
+                      hasSelectAll={false}
+                      overrideStrings={{ selectSomeItems: "State" }}
+                    />
+                    {/* <Multiselect
```
jgillis1, line 96: Could you remove the commented out pieces that are not being used here?

jgillis1 (changes requested):

frontend/react/src/components/sections/homepage/CMSHomepage.js
TGriesemer: I don't know if we "need" them but since they were already there I figured I would leave them. I will remove them and if we need them back we can add them in.

frontend/react/src/components/sections/homepage/CMSHomepage.js
TGriesemer: same as above.

jgillis1 (approved):

PR #1027, opened 2021-02-12
Automate DMS event for data transformation

ayoneeyee:


dwhitestratiform (approved):

dwhitecl:
@jgillis1 this code has no functional repercussions after being merged. this is all handling the data replication and data messaging and is applied using terraform locally. The code is checked in for tracking and historical purposes. Berry and Ayo confirmed that the lambda runs after a data migration and performs the task that Berry was previously running by hand. This is fine to merge to master.

PR #1032, opened 2021-02-23
Dev carts ses

mskalkar:
New service : dev-carts-ses


berryd (commented):
This will fire when it is deployed, do we want that?
I don't see the config for running this on a schedule.  How will this be invoked in the future?

mskalkar:
> This will fire when it is deployed, do we want that?
> I don't see the config for running this on a schedule. How will this be invoked in the future?

This Lambda function will be revoked by CloudWatch event scheduler which will be created later. The function will send email only if the db user password is about to expire (15 days or earlier).

mskalkar:
Added cron schedule to the script

berryd (approved):

PR #1098, opened 2021-09-07
upgrade from postgres 9.6 to postgres 12.7

dwhitestratiform:
**Purpose**
Postgres 9.6 end of life is in January 2022. AWS will automatically upgrade the RDS instance for you if you dont before then. We would like to upgrade proactively. This PR accomplishes that. 

**Linked Issues to Close**
https://qmacbis.atlassian.net/browse/OY2-10571

**Approach**
This PR simply updates the postgres terraform code. Upping the version to 12.7, allowing for major version upgrade, and also applying the upgrade immediately. If you do not do this, it will upgrade during a maintenance window. We also need to create a new parameter group. Being that the parameter group is based on per family... modifying the existing code block created a weird scenario where it would try and delete the existing parameter group but could not because its actively attached to an instance. 

This PR was tested on a clean deploy as well as on an existing 9.6 environment ensuring the deploy and upgrade went through cleanly. 

****** Important ****
DB upgrades can come with some risk. Its recommended snapshotting the RDS prior to merging/deploying to ensure we have the latest backup should something not go cleanly.** 

Update: snapshotting causes an issue. The upgrade tries deleting the 9.6 option group and cannot complete because the option group is attached to the manual snapshot. 

02/22/22 update:
deploying staging and prod at a later date revealed that the option group attached to each of those DB's was not the default option group but rather something that was being managed by terraform. This was a hard blocker deploying because it was trying to delete the option group attached to the DB and couldn't. The workaround was to modify the option group of the RDS instance for staging and prod by using the aws cli and running 

`aws rds modify-db-instance --db-instance-identifier <your_database_identifier_here> --option-group-name default:postgres-12 --apply-immediately`

**Learning**
postgres 9.6 cannot be upgraded to 13.* directly ... thats why 12.7 was chosen 

**Other**
This code should go in after the pipeline is clear of current in progress/testing work is released to production. Recommended limiting the next production release to strictly this  if possible to limit the scope of changes.
This change will incur downtime so that should be considered ahead of merging to produciton.


zachclarity (approved):
looks good

PR #1132, opened 2021-10-03
Dev nonokta Login

hasanfar:
Endpoint: http://d3negaa2ss2lz1.cloudfront.net/
Purpose: Enable non-Okta login for CARTS application


zachclarity (commented):
I could not get it to work.  Attached video.

zachclarity (changes requested):

zachclarity (approved):

zachclarity (approved):

zachclarity (changes requested):
Would provide a test with real data ?  I am trying to use it with real data and do not see values for forms for some reason.

zachclarity (approved):

PR #1159, opened 2021-10-14
Dev lambda

hasanfar:
Purpose: CARTS AVScan Lambda function policies should prohibit public access

Story: https://qmacbis.atlassian.net/browse/OY2-12508

Details: Added account id to the arn to restrict access only to the given account
image


gitsierra (approved):
👍

PR #1178, opened 2021-10-22
Carts/Seds Integration

13bfrancis:


benjaminpaige (approved):
maybe just check to see if you can get away with removing rejectUnauthorized: false. Both otherwise it looks great brian!

benjaminpaige (approved):

PR #1189, opened 2021-10-28
add N/A to section 1 Part 3&4

zachclarity:
https://qmacbis.atlassian.net/browse/OY2-13184


benjaminpaige (approved):
these are fun changes that we will have to adjust after the reporting season is over.

PR #1192, opened 2021-10-29
Update readme for onboarding

benjaminpaige:
Update readme for onboarding
update percentage component to allow for single decimals after number without following number.


jdinh8124 (approved):
Looks good to me, thanks for getting me set up too

PR #1193, opened 2021-11-01
New Way to Pass Custom Username and Password (Works for Both Terraform and Serverless)

13bfrancis:
The purpose of this PR is to allow for custom database login information to be passed into each environment. This was previously implemented, but there was an oversight. That being, custom credentials were not passed to serverless. This PR will fix that issue with minor changes to the way that we will be passing custom credentials. This page in the WIKI will be updated to reflect the new way of handling it.

https://github.com/CMSgov/cms-carts-seds/wiki/Password-Changes-for-the-Postgres-Database


mdial89f (approved):
The S3 bit aside (I'm out of the loop on that change)... the changes look good.

The docs aren't updated yet but Approving with the understanding that somewhere they will be updated.

LGTM

hasanfar (approved):

PR #1196, opened 2021-11-07
OY2-13439  Dynamically calculate % change for S2P1 in Carts

abelzer212:
This code should be temporary, and is absolutely the worst way to do this kind of change, but it's a one off, and my intention is to change it as soon as we do the CARTS rewrite


benjaminpaige (approved):
No worries. We've all done something similar in this project. If it works, im 👍

gitsierra (approved):
👍

13bfrancis (changes requested):

13bfrancis (approved):

PR #1748, opened 2021-12-13
OY2-14328: Fix section 5

gitsierra:
https://qmacbis.atlassian.net/browse/OY2-14328

Some of the data in the database had numbers with commas in them as strings so the frontend wasn't able to properly handle the parsing of the commas when using those numbers in formulas. This addresses that issue

### DIFF
--- a/frontend/react/src/util/synthesize.js
+++ b/frontend/react/src/util/synthesize.js
@@ -135,6 +135,7 @@ const formula = (targets, providedFormula, precision) => {
       try {
         targets[i] = targets[i].replace(",", "");
       } catch {
+        //do nothing
       }
       if (!Number.isNaN(targets[i]) && targets[i] !== "") {
         const replaceValue = new RegExp(`<${i}>`, "g");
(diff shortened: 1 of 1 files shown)


frontend/react/src/util/synthesize.js
```diff
@@ -132,6 +132,11 @@ const formula = (targets, providedFormula, precision) => {
     // Loop through formula as an object
     Object.keys(manipulatedFormula).forEach((i) => {
       // Check if value has a string value
+      try {
+        targets[i] = targets[i].replace(",", "");
+      } catch {
+        //do nothing
+      }
```
benjaminpaige, line 139: is this asynchronous? just wondering why you put it in a try catch?

jdinh8124 (approved):

frontend/react/src/util/synthesize.js
gitsierra: some of the targets it loops through are null/undefined/NaN so it throws an error when .replace gets called. I tried putting those into an if clause but it was being manipulated in an odd manner and it was still getting through. I thought this would be more concise as well.

benjaminpaige (approved):

PR #1894, opened 2022-02-18
Temporary fix to skip the data deploy

13bfrancis:
This will be a temporary fix while we look for the problem with options groups not being able to be deleted. For the time being I have added an environment variable (skip_data_deploy) which will skip the data deploy when set to true. It is currently being defaulted to true so unless you want to deploy the data layer no changes are required.


dwhitestratiform (approved):
we should rip this out after we get the 508 work released and resolve the postgres option group issue

PR #1923, opened 2022-03-01
Allow fake user authentication for local development

bodnarbm:
Fake user authentication is handy during local development to allow switching between various user roles. This change brings back the Fake user authentication scheme already supported by the frontend, but only for local environments (i.e. when running the local and dev docker compose files).

This can be tested locally by running the following within the `frontend` folder

```
docker-compose -f docker-compose.dev.yml up -d --force-recreate --build
```

Then after connecting to `http://localhost:81`, you should be able to use the various login types again on the left side of the screen.

This currently only brings this back for local runs of the app, and not the deployed dev version.


bodnarbm:
@13bfrancis this is the branch I was mentioning earlier to bring back the fake user authentication options for when we are developing locally.

abelzer212 (approved):
This all looks good, I'll wait for one more person to approve it and then pull it in

gitsierra (approved):
Looks good! good add with the local settings

PR #1924, opened 2022-03-01
Fix broken json schema for backend sections

bodnarbm:
The generate fixtures command was not executing locally due to an syntax error in the backend-section.schema.json. This fixes that error, allowing for fixtures to be generated again.

Test Plan
========

Ran the docker-compose dev stack, and checked the output of the api_postgres service to ensure there were no more errors related to the schema validation. Additionally checked that the fixtures folder now contains the various state level fixtures.

```
cd frontend
docker-compose -f docker-compose.dev.yml up --force-recreate --build -d
docker-compose -f docker-compose.dev.yml logs api_postgres --follow
ls api_postgres/fixtures | head
```

### DIFF
--- a/frontend/api_postgres/utils/section-schemas/backend-section.schema.json
+++ b/frontend/api_postgres/utils/section-schemas/backend-section.schema.json
@@ -185,14 +185,16 @@
         },
         "questions": {
           "type": "array",
-          "items": [
-            {
-              "$ref": "#/definitions/question"
-            },
-            {
-              "$ref": "#/definitions/repeatables"
-            }
-          ],
+          "items": {
+            "anyOf": [
+              {
+                "$ref": "#/definitions/question"
+              },
+              {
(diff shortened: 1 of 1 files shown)


bodnarbm:
Python format check seems to be failing because there are no changes to `.py` files and so nothing is being passed to `black` for checking. I suggest we address that in a separate PR.

abelzer212 (approved):
I see that one of the checks around Python style is failing, I think that was already happening though. Everything else here looks good to me. I just want one other team member to approve before I pull though

bodnarbm:
Thanks @abelzer212. Yeah, the python style checks look to be failing anytime there is a PR without any changed python files. This is an issue with the `git diff` command passing an empty string for the black's expected SRC parameter. I went ahead and did another PR #1926 to fix that up.

gitsierra (approved):

PR #1926, opened 2022-03-02
Fix python format check in GH actions

bodnarbm:
GitHub Actions were failing for the python format check if the PR contained no changed python files. This is because black requires some argument to the SRC and the git diff was returning an empty string.

This changes to use the provided GitHub action by psf (the makers of black) and runs on all python files, not just the changed ones. black is fast enough that we can run it over all python files.


TheDanMiller (approved):

gmrabian (approved):
lgtm, passed the github checks using the new steps

karla-vm (approved):

PR #1927, opened 2022-03-02
Use state list when creating new state statuses.

bodnarbm:
When the admin user triggers generation of a new form year, the StateStatus objects were only being created for States that had a prior StateStatus object. This would result in no StateStatus objects being created in a brand new environment, like local.

Here, we switch is enumerating over the States themselves to ensure that a StateStatus object exists for every state for the new form year.

Test Plan
=========

Tested this locally on a fresh local stack by logining in as an admin, generating new base forms for FFY 2021, and then checking that there is now a FFY 2021 for each state.

```
cd frontend
docker-compose -f docker-compose.local.yml down -v --rmi local
docker-compose -f docker-compose.local.yml up -d
# Wait for services to be fully up, then connect to http://localhost:81
```


BearHanded (approved):

gmrabian (approved):
very logical

karla-vm (approved):

PR #1933, opened 2022-03-03
Bring back fake user authentication from #1923

bodnarbm:
After digging in a bit yesterday, we determined that the deployment failures were not related to #1923, but instead to the deployment environment itself. While that is being addressed, I wanted to put up this PR to undo #1930's revert of #1923.

cc/ @13bfrancis and @benjaminpaige

This reverts commit 9170f3c5532a30fd28e3f38a8996ed007c01bbad.


bodnarbm:
Addressed by #1938

PR #2012, opened 2022-03-17
Dev 508 - Explicityly add arn to policy

13bfrancis:


frontend/react/src/App.js
```diff
@@ -6,7 +6,7 @@ import WrappedSecurity from "./wrapSecurity";
 
 // Add comment to force build frontend (will revert)
 function App() {
-  // This is where the react app gets kickstarted!!
+  // This is where the react app gets kickstarted!!!
```
benjaminpaige, line 9: nice!!!

benjaminpaige (approved):

frontend/react/src/App.js
13bfrancis: Wonder how many exclamation points we will end up with

PR #2013, opened 2022-03-17
[MDCT-51] Patch Django

juliuscaesar:
https://qmacbis.atlassian.net/browse/MDCT-51

Upgrades Django from 2.2.13 to the v2 LTS, 2.2.27. 

I verified Django upgraded to new version, no dependencies broke, and everything seemed normal on my limited (no reports) local setup.

### DIFF
--- a/frontend/api_postgres/requirements.txt
+++ b/frontend/api_postgres/requirements.txt
@@ -1,4 +1,4 @@
-Django==2.2.13
+Django==2.2.27
 gunicorn==19.9.0
 psycopg2-binary==2.8.3
 djangorestframework


karla-vm (approved):

TheDanMiller:
@juliuscaesar or @karla-vm did either of you go through the workflow after the upgrade to verify states can still do their thing and co/bus user can still review

gmrabian (approved):
Tested locally, clicked through the flow with each user and it all seems good to me.
Also verified the logs indicate it's using the right version:
`Django version 2.2.27, using settings 'carts.settings_local'`

karla-vm:
> 

Yes, the flow is looking good on my local! Nothing broken and the correct Django version is showing up.

PR #2018, opened 2022-03-21
Update react-scripts version from 3.4.1 to 5.0.0

karla-vm:
https://qmacbis.atlassian.net/browse/MDCT-50

Updating react-scripts from 3.4.1 to 5.0.0, tested and verified on local environment.


gmrabian (commented):
Alright I'm a JS noob so help me out - the package.json has the new version (5.0.0) and package-lock.json has 3.4.1. Does it matter?

braxex (changes requested):
As discussed, let's do `npm install --save --save-exact react-scripts@5.0.0`, then do some more local regression testing to make sure everything's working. 

Also adding a `.nvrmc` pinned to `14.4.0` would be cool. 👍

frontend/react/.nvmrc
```diff
@@ -0,0 +1 @@
+14.0.0
```
bodnarbm, line 1:
```suggestion
14.4.0
```

I think to match the [Dockerfile](https://github.com/CMSgov/cms-carts-seds/blob/424221167fbc5f08103ce4bc0fe4175ace91e83f/frontend/react/Dockerfile#L3) and the [Jenkinsfile](https://github.com/CMSgov/cms-carts-seds/blob/424221167fbc5f08103ce4bc0fe4175ace91e83f/.jenkins/Jenkinsfile.master#L39).

karla-vm: Done! Thanks for the heads up.

braxex (approved):
Looking good. Thanks!

juliuscaesar (approved):

bodnarbm (changes requested):
@karla-vm, following up on our debugging, I think something went wrong with our package lock file. Between all these commits. I recreated the work that we did (with some minor tweaks that I don't think matter) in another branch [react-upgrade](https://github.com/CMSgov/cms-carts-seds/commit/56b28d229cc33128a1927da7be0f16f01690d61f). For whatever reason this one doesn't have that dropdown issue that we were experiencing. Best I can tell, the only real difference between these two branches is the package-lock.json.

Adding the request changes to block merge until we resolve.

PR #2022, opened 2022-03-21
Dev 508 Change Allowed Resources

13bfrancis:


kenrickveit (approved):
Approved, we don't want this hardcoded anyway as the account and api id will very by env. 
If we need to move away from wildcards at some point, we should pull this from an SSM param

benjaminpaige (approved):

PR #2025, opened 2022-03-21
Skips the main branch from deploy. Keeps it intact otherwise for veri…

TheDanMiller:
…fication


braxex (approved):
👍

BearHanded (approved):

gmrabian (approved):

gmrabian:
(I know I already approved...)
IMO we could just remove that deploy block you commented out. If we are going to write our own we can always reference it in the history but no need to leave it. If you think ours will be almost identical then maybe it's worth leaving 🤷

BearHanded (approved):

PR #2026, opened 2022-03-22
MDCT-35 - Tech Cleanup - Bring in CARTS UI

BearHanded:
## Purpose

https://qmacbis.atlassian.net/browse/MDCT-35

## Approach

Brought in the CARTS UI into the stack. Currently deploys the UI with the rest of the pieces, still need to replace the auth portion to have an interactive experience


gmrabian (commented):
Should we remove the react imgs? ui-src/public/logo*.png

Otherwise I think it's fine. Just porting stuff over. I'm guessing a lot of the removed stuff was in the quickstart but not CARTS?

gmrabian (dismissed):
Ran locally, can see the user select page

gmrabian (approved):

karla-vm (approved):
I'm able to see the CARTS UI locally, looking good.

PR #2082, opened 2022-03-25
Fix bottom print button

13bfrancis:
# Description

This will fix the bottom print button to generate a 508 compliant pdf.


TheDanMiller (approved):
This is fine. It hurts my brain that this does not exist as a reusable component and has to be updated twice -- but this will get it done.

13bfrancis:
It hurts me too. There are days I wish I could go through and refactor all of this stuff 😂

PR #2084, opened 2022-03-29
MDCT-40: Miscellaneous main branch frontend cleanup

braxex:
# Description

This PR just has some low-level cleanup in it. There were several areas of the frontend that needed a bit of cleanup. There's also admittedly more work that can be done, but for now this provides a cleaner baseline to build on.

## Things That Changed
- Renamed several files
- Fixed a bunch of linting errors & warnings
- Moved around some styling to make it cleaner

## Things I'm Not So Sure About
- The `react-hooks/exhaustive-deps` stuff is tricky. I'll let you do your own research and come to your own conclusions, but at the end of the day I think disabling the linting on this currently working code is the best move. I started down the path of digging into the depths of each of the areas where this was an issue, but it quickly became obvious that would be a bit of a waste of time and potentially open us up to some unnecessary risk. My goal in fixing a bunch of these warnings that were showing up in the console was to limit the amount of noise so folks can debug and identify new issues easily. Open to hearing thoughts.

## Things To Be Addressed In Future PRs
- There are a bunch of styling rules that I've moved to `_overrides.scss`. They should be scoped more closely, but I don't want to do that until we have auth finalized and everything working right so we can ensure the needed styling changes propagate correctly. 
- There is so much noise in the console, most of it due to `PropTypes` shenanigans, though the `redux-logger` is pretty loud too 😂. If we have time, we should go through and clean up the prop-related stuff.

## How to test

Nothing to really test at this point.

## Dependencies 

No changed dependencies.

# Get it done

## Code authors checklist
- [X] I have performed a self-review of my code
- [X] I have added thorough tests. [Testing Doc](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research)
- [X] Necessary analytics were added, or no new analytics were required
- [X] I have made corresponding changes to the documentation
- [X] I have assigned the PR to myself

- [x] I have done the lgtm review 

## Assignee 


gmrabian (dismissed):
Code level changes look good to me. I don't know as much about SCSS but this looks like mostly cleanup. I ran locally and worked fine. 

About the linting thing - I read about it a bit. Is the thing we're avoiding just listing a bunch of dependencies in that block?

braxex:
>About the linting thing - I read about it a bit. Is the thing we're avoiding just listing a bunch of dependencies in that block?

@gmrabian Not exactly. The real issue that's being called out is that the `useEffect` hook has dependencies that are not explicitly listed in that dependency array. That array exists to let the hook know what objects to "watch" (it's not actually "watching" per se but that's another topic), and when any of those declared deps changes, the hook will run again. Declaring all of the dependencies of the code in that hook is helpful because it avoids bugs by ensuring the hook isn't running with stale data. And so we have the `exhaustive-deps` linting rule. But declaring all of the dependencies as the solution often causes more problems. For instance, functions in the hook being referenced in other places, but which also references and update component state (sadly the case with a lot of these). Adding those pieces of state as deps would cause infinite loops. 😕 

There are a few potential fixes, but I kinda want to wait until we have everything hooked back up and running to mess with it any further because I'd hate to think we'd solved them only to find out we've broken some functionality we don't yet understand. Considering the current state of some of these hooks, unexpected side-effects are likely. My hope is sometime toward the end of the year we can circle back to these with a more comprehensive view of what's going on across the repe -- then we can evaluate what the best rewrites/solutions will look like. 🤷

gmrabian (approved):

BearHanded (changes requested):

BearHanded (approved):

PR #2189, opened 2022-04-06
MDCT-97: Jest Setup

karla-vm:
# Description

Closes out [MDCT-97](https://qmacbis.atlassian.net/browse/MDCT-97). In this PR, Jest is setup for the `ui-src` directory to start unit testing the frontend. We still don't have any unit tests for the API because we haven't written any endpoints as of yet. However, there is a sample unit test for a frontend component under `services/ui-src/src/__tests__ ` for `UserProfile.js`.

## How to test

- `cd services/ui-src`
- `nvm use`
- `yarn install`
- `yarn test`

## Dependencies 

Run `yarn install` on your terminal.

# Get it done

We get to write Jest tests!

## Code authors checklist

## Assignee 


gmrabian (dismissed):
lgtm, works locally - remember `nvm use`

karla-vm:
> lgtm, works locally - remember `nvm use`

Thank you! I added that to the instructions for testing.

services/ui-src/package.json
```diff
@@ -53,7 +53,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
```
bodnarbm, line 56: Do we need to change away from `react-scripts test`? react-scripts already uses `jest` as it's test runner. Was there something it was doing that was problematic for us?

bodnarbm (changes requested):

karla-vm:
@braxex can I get a quick set of eyes on this to merge into main?

services/ui-src/src/__tests__/UserProfile.test.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { shallow } from "enzyme";
+import "../setupTests";
```
bodnarbm, line 3: We actually want to do this in the jest config itself, to ensure this is ran with every test.

It would be something like

```
"setupFilesAfterEnv": [
      "<rootDir>/src/setupTests.js"
    ],
```

services/ui-src/src/__tests__/DateRange.test.js
```diff
@@ -0,0 +1,93 @@
+// TODO: Update testing suite with DynamoDB endpoint (currently points to PostgreSQL endpoint)
```
bodnarbm, line 1: Can you explain more about this todo? None of the unit tests should actually hit the endpoint so this should not matter for component tests.

If this is because of `window.env.API_POSTGRES_URL` not being set at all, that can be handled by adding something like

```
global.window.env = {
  API_POSTGRES_URL: "fakeurl",
};
```
in the setupTest.js file.

services/ui-src/src/__tests__/DateRange.test.js
```diff
@@ -0,0 +1,93 @@
...
+
+describe("placeholder test", () => {
+  test("test suite runs", () => {
+    expect(1 + 2).toBe(3);
+  });
+});
+
+/*
```
bodnarbm, line 9: Let's make sure we add an issue to fix all these tests. Don't want them to be forgotten about later.

bodnarbm (changes requested):
Few changes requests throughout and noted in the comments. Some more global comments are

- The yarn.lock files generated by newer versions of yarn (yarn 2+) are not compatible with yarn 1. Can we please rerun this with a yarn 1.
- We will want to remove the [.yarn/install-state.gz](https://github.com/CMSgov/cms-carts-seds/pull/2189/files#diff-08726f0f086814efdbc11017d6e0102b08abac2d45d6078d359ca189d5989462) from the changes.
- I believe the pattern being followed in qmr and mcr is for tests to live next to the component they are testing. Thus, I think we should move all the files out from under `__test__` and instead place them next to each component. See https://github.com/CMSgov/cms-mdct-qmr/blob/4bc8235f873eb7b25c35db6922d94f36da8488da/services/ui-src/src/components/Header/index.test.tsx as an example

.gitignore
```diff
@@ -15,3 +15,4 @@ tests/cypress/videos
 tests/cypress/screenshots
 tests/cypress/package-lock.json
 package-lock.json
+.yarn/cache/
```
bodnarbm, line 18: nit: maybe remove this line?

bodnarbm (dismissed):
LGTM. Minor nit on gitignore.

services/ui-src/babel.config.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  presets: [
+    [
+      "@babel/preset-env",
+      {
+        targets: {
+          node: "current",
+          edge: "17",
+          firefox: "60",
+          chrome: "67",
```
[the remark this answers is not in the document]
bodnarbm: We want want to target lower versions of Chrome though as users may not be on the latest.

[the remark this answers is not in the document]
gmrabian: ah that makes sense. Thanks

.gitignore
karla-vm: I missed that one! Removed it in the latest commit.

gmrabian (approved):

bodnarbm (approved):

PR #2198, opened 2022-04-08
Feature/mdct 154 section titles prince

TheDanMiller:
# Description

This is about pdf bookmarks as an output of the prince service. PDF bookmarks are generated based on the structure of the application and the entire thing was written using h2. To remediate I did the following
1. Restructured the HTML to use a hierarchy
2. Updated styles so that look and feel did not change
3. Deleted a bunch of dead code
4. Fixed a spelling error
5. Refactored some really bad code in Section.js

## How to test

Hook up your front end to dev backend 
Print the PDF
Verify the bookmarks are better

## Code authors checklist
- [x] I have added thorough tests. (no changes) [Testing Doc](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research)

## Assignee 


frontend/react/src/components/fields/Objectives.js
```diff
@@ -62,9 +62,9 @@ const Objectives = ({
       </Accordion>
 
       <div className="section-footer">
-        <h3 className="question-inner-header">
+        <span className="question-inner-header span-pdf-no-bookmark">
```
gmrabian, line 65: for my own learning, does html acknowledge this as having two classes? Thus the new `span-pdf-no-bookmark` properties get applied here? Just curious why some only have `span-pdf-no-bookmark` and some have an additional class

gmrabian (approved):
Lookin good! Verified locally against master and the bookmarks are much cleaner. HTML, etc. changes look good, would need a second opinion on those. Only other thing is double check with Kim or whoever that this didn't worsen any other accessibility stuff (I'm sure it's fine)

Screen Shot 2022-04-08 at 9 40 16 AM Screen Shot 2022-04-08 at 9 40 10 AM

frontend/react/src/components/fields/Objectives.js
TheDanMiller: So in this instance I added span-pdf-no-bookmark to the existing class to not change what it was doing. It does acknowledge the two classes though.

BearHanded (approved):

PR #2209, opened 2022-04-11
MDCT-105 - State View

BearHanded:
# Description

Allow the state user to view all reports for their state.
Additionally provides some functionality for the other user types to view all reports, and retrieve some dummy data from the /state endpoint so those views don't crash and burn.

Sets up seed StateStatuses for the local environment for Alabama. This is easy enough to iterate on and expand if other entries are needed.

## How to test

* Running locally, and logging in as StateUser2 should allow you to see Statuses for Alabama, and only Alabama. 
* The links for View/Edit should exist based on the Report Status.
* A modified date has been added for each entry for the stateUser

## Dependencies 

The .env_example file in the root level has a few new props that you will want to copy into your .env to be able to run the app.

# Get it done

Look through, Test and Approve

## Code authors checklist

## Assignee 
- [x] I have closed the PR after the review and necessary changes (squashing preferred)


services/ui-src/src/types.js
```diff
@@ -3,5 +3,25 @@ export const UserRoles = {
   STATE: "mdctcarts-state-user",
   HELP: "mdctcarts-help-desk",
   BO: "mdctcarts-bo-user",
-  CO: "todo-co",
+  CO: "mdctcarts-co",
+};
+
+export const REPORT_STATUS = {
```
gmrabian, line 9: as I recall we narrowed these down to just not-started, in-progress, and certified, right?

gmrabian (dismissed):

services/ui-src/src/types.js
BearHanded: yes, however they do exist for historical reports, so it might be best to keep some reference to them around

braxex (dismissed):

services/app-api/handlers/stateStatus/get.ts
```diff
@@ -0,0 +1,32 @@
+import { getUserCredentialsFromJwt } from "../../libs/authorization";
+import handler from "../../libs/handler-lib";
+import dynamoDb from "../../libs/dynamodb-lib";
+import { UserRoles } from "../../types";
+import { convertToDynamoExpression } from "../dynamoUtils/convertToDynamoExpressionVars";
+
+export const getStateStatus = handler(async (event, _context) => {
+  const user = getUserCredentialsFromJwt(event);
+
+  if (user.role === UserRoles.STATE && user.state !== undefined) {
```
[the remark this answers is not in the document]
BearHanded: state could be undefined if a state user was configured incorrectly, but it should capture other falsy values, good call

braxex (approved):

gmrabian (approved):
learned something new

PR #2258, opened 2022-04-20
Add h1 and pdf title

gmrabian:
# Description

Our PDFs were failing two accessibility findings: h1 and document title. This resolves both.

- Broke out the h1 title into its own component so it can be used independently of the PageInfo component.
- Modified Redux state to pull the form's state name from the path (applies to viewing/editing form)
- Added a title to the printable report as well as the h1 object. Pulls state name and form year from query params (Since this opens in a separate tab couldn't reuse state)
- Created unique styling for title appearing independently vs with the PageInfo (different left margins)
- Added a query param check to the redux form year update (fixes bug where a form from 2021 displays with year 2020 on the print page)

If you have suggestions for improving this, please let me know!!

[Ticket](https://qmacbis.atlassian.net/browse/MDCT-153?atlOrigin=eyJpIjoiMmIyZGEyNjMwY2FjNDA1MWE1N2I1ZDM5ZDk4MWQ1ZjMiLCJwIjoiaiJ9)

Any tests I can add to this/will this stay the same in v3?

## How to test

1. Change into the react directory under `frontend/react`
2. `nvm use 14.4`
3. Run the following command to create a local environment that points to the deployed dev api `API_POSTGRES_URL=https://api.mdctcartsdev.cms.gov/ ./env.sh && cp env-config.js public/`
4. Start react locally with `npm run start`
Then you can login and do everything from localhost:3000 (incognito recommended)

Things to look for: State name shows up properly when switching between reports in admin view.
When I select print entire form, the correct title displays on the preview. When I click print on the preview, the document has a title in the format `{State Name} CARTS FY{XXXX} Report`

## Code authors checklist
- [x] I have performed a self-review of my code
- [ ] I have added thorough tests. [Testing Doc](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research)
- [x] Necessary analytics were added, or no new analytics were required
- [ ] I have made corresponding changes to the documentation
- [x] I have assigned the PR to myself

## Reviewers Checklist (Two different people)
- [ ] I have done the deep review and verified the items in the above checklist are g2g
- [ ] I have done the lgtm review 

## Assignee 
- [ ] I have closed the PR after the review and necessary changes (squashing preferred)

PR #2290, opened 2022-04-25
Fmap table

gmrabian:
# Description

FMAP data from SEDS was not showing up for admins. Reason is admins don't have state info in their profile, which is where a lot of things fetch info from. Switched the table value loader to load from global state (already set up this situation in a previous [PR](https://github.com/CMSgov/cms-carts-seds/pull/2258) - basically comes down to [this line](https://github.com/CMSgov/cms-carts-seds/blob/master/frontend/react/src/components/Utils/InvokeSection.js#L16) that assigns a global state name upon loading the form for that state.) Now any user role will be able to see and print the same table.

Also confirmed separately with Daniel Elias that 2023 does not have data in SEDS yet, so below image (from dev) is best we can do.
Screen Shot 2022-04-25 at 1 10 55 PM

[Ticket](https://qmacbis.atlassian.net/browse/MDCT-75?atlOrigin=eyJpIjoiMTRmODc2OTc5ZjY3NDc5OWFiOTZjZjVkYWFjMDM3ZTMiLCJwIjoiaiJ9)

## How to test

1. Change into the react directory under `frontend/react`
2. `nvm use 14.4`
3. Run the following command to create a local environment that points to the deployed dev api `API_POSTGRES_URL=https://api.mdctcartsdev.cms.gov/ ./env.sh && cp env-config.js public/`
4. Start react locally with `npm run start`
Then you can login and do everything from localhost:3000 (incognito recommended)

Open Alaska 2021, go to section 5, view table 3 and confirm it looks similar to screenshot (less concerned about the numbers, just the populated cells.)

## Code authors checklist

## Assignee 

### DIFF
--- a/frontend/react/src/util/synthesize.js
+++ b/frontend/react/src/util/synthesize.js
@@ -176,10 +176,17 @@ const sum = (values) => {
 };
 
 const lookupFMAP = (state, fy) => {
-  if (state.allStatesData && state.global.stateName) {
-    const stateData = state.allStatesData.filter(
-      (st) => st.name === state.global.stateName
-    )[0];
+  if (state.allStatesData && (state.global.stateName || state.stateUser.abbr)) {
+    let stateData = "";
+    if (state.stateUser.abbr) {
+      stateData = state.allStatesData.filter(
+        (st) => st.code === state.stateUser.abbr
+      )[0];
+    } else {
+      stateData = state.allStatesData.filter(
+        (st) => st.name === state.global.stateName
+      )[0];
+    }
     const fmap =
       stateData?.fmap_set.filter((year) => year.fiscal_year === +fy)[0]
         ?.enhanced_FMAP || NaN;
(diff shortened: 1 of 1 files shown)


frontend/react/src/util/synthesize.js
```diff
@@ -176,10 +176,9 @@ const sum = (values) => {
 };
 
 const lookupFMAP = (state, fy) => {
-  if (state.allStatesData && state.stateUser) {
-    const stateAbbr = state.stateUser.abbr;
+  if (state.allStatesData && state.global.stateName) {
```
BearHanded, line 179: this block should check for a stateName on either the state.global.stateName or the state.stateUser.abbr. The value is stored in one or the other of those locations based on the user type.

BearHanded (changes requested):

BearHanded (approved):

BearHanded:
Approved and tested with state user

TheDanMiller (approved):

PR #2291, opened 2022-04-25
cleanup local console warnings

gmrabian:
# Description

Push a few versions and make small adjustments to slim down the logs when running `./dev local`

Reduces startup logs from 314 lines to 235.

Does not cover every issue in this [Ticket](https://qmacbis.atlassian.net/browse/MDCT-164?atlOrigin=eyJpIjoiMDQ4ZDc3Zjg5ZTFiNGRlOTg3OTYyMGNmM2IxZjQxNjciLCJwIjoiaiJ9) because setting some local defaults requires a deeper look imo. I don't want to set too many defaults that could trickle up to deployed versions.

## How to test

- `nvm use`
- `./dev local`

Verify the app still works as expected

Enjoy the shorter log output 😎

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "serverless-dotenv-plugin": "^3.0.0",
     "serverless-iam-helper": "github:cmsgov/serverless-iam-helper",
     "serverless-idempotency-helper": "github:cmsgov/serverless-idempotency-helper",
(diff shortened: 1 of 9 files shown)


braxex (approved):
Pulled and confirmed that things in the terminal indeed look 👌

BearHanded (approved):

PR #2321, opened 2022-04-29
DynamoDB Best Practices and Unit Tests

juliuscaesar:
# Description

Quick PR for DynamoDB best practices and unit tests.

Changes:
* Uses `query` instead of `scan` for the `section` endpoints
* Updates the `section` unit tests accordingly
* Adds unit tests for `state status`
* Adds backend unit tests to github actions

## Code authors checklist

## Assignee

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,4 +1,4 @@
-name: Pre-commit
+name: Pull-request
 
 on: [pull_request]
 
@@ -11,17 +11,34 @@ jobs:
       - uses: pre-commit/action@v2.0.2
       - uses: actions/setup-node@v3
         with:
-          node-version: '14'
+          node-version: "14"
+  jest-frontend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v2.0.2
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "14"
       - name: Prepare for Jest (frontend)
         run: yarn install --frozen-lockfile
         working-directory: ./services/ui-src
       - name: Jest (frontend)
         run: yarn test
         working-directory: ./services/ui-src
+  jest-backend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v2.0.2
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "14"
       - name: Prepare for Jest (backend)
         run: yarn install --frozen-lockfile
         working-directory: ./services/app-api
       - name: Jest (backend)
         run: yarn test
         working-directory: ./services/app-api
-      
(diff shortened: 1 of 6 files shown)


gmrabian (dismissed):
Tests passing. I don't know much about scan but I like the use of query - that's what I've usually seen on the backend. 

Nice work!

.github/workflows/pre-commit.yml
```diff
@@ -12,10 +12,16 @@ jobs:
...
           node-version: '14'
-      - name: Prepare for Jest
+      - name: Prepare for Jest (frontend)
         run: yarn install --frozen-lockfile
         working-directory: ./services/ui-src
-      - name: Jest
+      - name: Jest (frontend)
         run: yarn test
         working-directory: ./services/ui-src
+      - name: Prepare for Jest (backend)
```
bodnarbm, line 21: non-blocking: not for this PR but on my mind because of this: We may want to start splitting up some of these steps into separate jobs within the same workflow, to allow taking advantage of parallelism as these tests become a bit slower.

services/app-api/handlers/section/get.ts
```diff
@@ -24,16 +23,14 @@ export const getSections = handler(async (event, _context) => {
...
     const params = {
       TableName: process.env.sectionTableName!,
-      ...convertToDynamoExpression(
-        {
-          year: parseInt(year),
-          stateId: state,
-        },
-        "list"
-      ),
+      KeyConditionExpression: "pk = :pk AND sectionId >= :sectionId",
```
bodnarbm, line 26: blocking question: Doesn't the javascript api allow you to specify only the partition key and just not specify the sort key at all?

Something like

``` typescript
KeyConditionExpression: "pk = :pk",
ExpressionAttributeValues: {
  ":pk": `${state}-${year}`,
},
```

services/app-api/handlers/section/get.ts
```diff
@@ -24,16 +23,14 @@ export const getSections = handler(async (event, _context) => {
...
+      ExpressionAttributeValues: {
+        ":pk": `${state}-${year}`,
+        ":sectionId": 0,
+      },
     };
 
-    const queryValue = await dynamoDb.scan(params);
+    const queryValue = await dynamoDb.query(params);
```
bodnarbm, line 33: This is probably not going to happen with our number of sections and the expected size, but since the query can return more than one response, best practice would be the check if there is another page (check if there is a `LastEvaluatedKey` in the require value) and then handle the other pages if they exist (here I assume we would just retrieve them all).

bodnarbm (changes requested):
Just some minor items on the get section query.

services/app-api/handlers/section/get.ts
juliuscaesar: I recalled getting an error back there being a mismatch in the key schema but it seems to be working without it now. Just pushed an update.

services/app-api/handlers/section/get.ts
juliuscaesar: I recalled getting an error back there being a mismatch in the key schema but it seems to be working without it now. Just pushed an update.

services/app-api/handlers/section/get.ts
bodnarbm: Changing this to non-blocking as the current section sizes look to be about 250-260KB and we can handle the additional paging later.

bodnarbm (approved):

gmrabian (approved):
I like the parallelism

PR #2346, opened 2022-05-05
MDCT-163 - Frontend console cleanup

gmrabian:
# Description

Progress on [ticket MDCT-163](https://qmacbis.atlassian.net/browse/MDCT-163?atlOrigin=eyJpIjoiMTcyNTFkNWIzZjg0NGE1OTkyZTcwNjAwNThkMWI3OWYiLCJwIjoiaiJ9). Still more errors and warnings in console, though just putting these in for now so as not to make a small thing a big PR.

Biggest change is cmsgov/design-system-* is [now deprecated in favor of design-system](https://design.cms.gov/startup/migrating-v2/). Mostly oneliners on this.

Small redux bumps and package changes for migrated extensions package.

Changed some props to be the right type, or not required when they aren't.

LMK what you think! I tried to split up the commits so we can drop one of these if they're not good. If I have time before the end of the sprint I'll hit a few more.

## How to test

Run locally, verify everything works, verify changes look ok in code.


karla-vm (approved):
My console looks _MUCH_ cleaner.

BearHanded (approved):
Looks better than before. There are still a bunch of react warnings that need to be cleaned up, but this is progress.

PR #2562, opened 2022-05-19
MDCT-252: Help Desk User View All CARTS Reports

karla-vm:
# Description

Closes [MDCT-252](https://qmacbis.atlassian.net/browse/MDCT-252). The new Help Desk user (functionally, an admin user) can view a list of all CARTS reports. These reports are now sorted by year, then alphabetically. For example:

`2022 Alabama`
`2022 Alaska`
`2022 California`
`2021 Alaska`
`2020 California`

Also, the "List Users", "Add Users", and "Generate Form Base Templates" buttons should be _gone_.

## How to test

`./dev local`

Log in as an admin user, check out how everything is sorted and those buttons don't exist anymore!

## Dependencies

N/A

# Get it done

N/A

## Code authors checklist

- [x] I have done the lgtm review

## Assignee

### DIFF
--- a/services/ui-src/src/components/sections/homepage/CMSHomepage.js
+++ b/services/ui-src/src/components/sections/homepage/CMSHomepage.js
@@ -62,6 +62,7 @@ const CMSHomepage = ({
   const clearFilter = () => {
     window.location.reload(false);
   };
+
   return (
     <div className="homepage ds-l-col--12">
       <div className="ds-l-container-large">
@@ -144,9 +145,10 @@ const CMSHomepage = ({
               </div>
               <div className="report-status">
                 {statuses
-                  .sort((a, b) => (a.stateCode < b.stateCode ? -1 : 1))
-                  .sort((a, b) => (a.year > b.year ? -1 : 1))
-                  .sort((a, b) => (a.lastChanged > b.lastChanged ? -1 : 1))
+                  // if there is a difference in year, sort by year first, otherwise, sort by state
+                  .sort(
+                    (a, b) => b.year - a.year || (a.state < b.state ? -1 : 1)
+                  )
                   .map(
                     ({
                       state,
@@ -157,7 +159,6 @@ const CMSHomepage = ({
                       lastChanged,
                     }) => {
                       return (
-                        // eslint-disable-next-line
                         <div>
                           {
                             // eslint-disable-next-line
--- a/services/ui-src/src/components/sections/homepage/CMSHomepage.js
+++ b/services/ui-src/src/components/sections/homepage/CMSHomepage.js
@@ -144,6 +144,8 @@ const CMSHomepage = ({
               </div>
               <div className="report-status">
                 {statuses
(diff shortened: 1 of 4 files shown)


gmrabian (dismissed):
works! excellent added:deleted line ratio

services/ui-src/src/components/sections/homepage/CMSHomepage.js
```diff
@@ -144,6 +144,8 @@ const CMSHomepage = ({
               </div>
               <div className="report-status">
                 {statuses
+                  .sort((a, b) => (a.stateCode < b.stateCode ? -1 : 1))
+                  .sort((a, b) => (a.year > b.year ? -1 : 1))
                   .sort((a, b) => (a.lastChanged > b.lastChanged ? -1 : 1))
```
bodnarbm, line 149: This last sort would resort everything based on the last time the data was changed regardless of the year and state code and should be removed.

services/ui-src/src/components/sections/homepage/CMSHomepage.js
```diff
@@ -144,6 +144,8 @@ const CMSHomepage = ({
               </div>
               <div className="report-status">
                 {statuses
+                  .sort((a, b) => (a.stateCode < b.stateCode ? -1 : 1))
```
bodnarbm, line 147: Is the sort intended to be by full state name or the state's code?

We display the state name to the user, so I believe the sort by state should be based on that. For example, for Alabama and Alaska, if we sort by name, Alabama comes first, but if we sort by state code, Alaska comes first.

services/ui-src/src/components/sections/homepage/CMSHomepage.js
bodnarbm, line 149: This last sort would resort everything based on the last time the data was changed regardless of the year and state code and should be removed.

services/ui-src/src/components/sections/homepage/CMSHomepage.js
```diff
@@ -144,6 +144,8 @@ const CMSHomepage = ({
...
+                  .sort((a, b) => (a.year > b.year ? -1 : 1))
```
bodnarbm, line 148: I think it may cause confusion later on to have this done via multiple sorts, as the reader of the code needs to remember that the the latest sort wins over the prior sorts and that browsers (from 2019 and later) have had stable sorts (which is the reason why we know that within a given year here that its sorted by the state information). My suggestion would be to do a single sort with a sort comparison function that first checks if they years differ and then returns based the year order, and if the years are the same then returns based on the state information.

Something like

```
# puesdocode
a, b:
  if a.year != b.year
    return comparison value based on year

  return comparison value based on state 
```

bodnarbm (changes requested):
In addition to the sort changes noted, @karla-vm, can you check when logging in as a user with the help desk role (like `george@test.com`). I believe the Home component currently routes help desk users to an unauthorized page.

bodnarbm (approved):
Thanks!

gmrabian (approved):

PR #2563, opened 2022-05-19
add nick

gmrabian:
# Description

Add Nick

Dan said we might have to wait until Nick officially has the GITHUB job role so we can hold this if need be


braxex:
The code is questionable, but I guess I'll reserve my many suggestions for after we get the green light. 😏

bodnarbm (approved):

braxex (approved):

PR #2565, opened 2022-05-20
MDCT 294 - Update Design System Dependency

ntsummers1:
# Description

Fixes [MDCT-294](https://qmacbis.atlassian.net/browse/MDCT-294). Real easy one in moving up the @cmsgov/design-system dependency in the ui-src/package.json file from 2.0.0 -> 3.4.0. Thankfully there are no breaking changes that we were using. The major version change was for a deprecation of a prop on the button component that we don't use (You can find more information about that [here](https://github.com/CMSgov/design-system/releases/tag/%40cmsgov%2Fdesign-system%403.0.0). I also don't see any other changes that affected our code. How convenient!

## How to test

Load up the project and explore each of the screens. Everything should look exactly the same with nothing different!

## Dependencies

@cmsgov/design-system dependency in ui/src upgraded from 2.0.0 -> 3.4.0

# Get it done

## Code authors checklist

- [ ] ~~I have added thorough tests. [Testing Doc](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research)~~
- [ ] ~~I have made corresponding changes to the documentation~~

## Assignee


services/ui-src/src/matchMedia.mock
```diff
@@ -0,0 +1,13 @@
...
+    matches: false,
+    media: query,
+    onchange: null,
+    addListener: jest.fn(), // deprecated
+    removeListener: jest.fn(), // deprecated
+    addEventListener: jest.fn(),
+    removeEventListener: jest.fn(),
+    dispatchEvent: jest.fn(),
+  })),
+});
```
ntsummers1, line 13: I don't really like how locally this test passes but seeing an error when running it through the Git Action. [This link](https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom) talks about the official method for fixing this issue when the Header was coming back with a matchMedia isn't implemented error. Weird.

ntsummers1: I also don't like the folder its in, but i kept it with the other local ui jest code. Should it be moved to a separate folder?

bodnarbm: If this was mock of a module or other component, it would go in a `__mocks__` folder either next to the module (if local) or next to the `node_modules` director (if an external dependency)

But since this is mocking something on the window object, I recomment placing it in the `services/ui-src/src/setupTests.js` file as that will be setup and included in all tests then.

bodnarbm (dismissed):
Suggestion on where to place the mock, but LGTM regardless.

services/ui-src/src/matchMedia.mock
gmrabian: +1 on setup tests

braxex: Yep, that would match what we've done in other repos as well. 👍

braxex (dismissed):

services/ui-src/src/matchMedia.mock
ntsummers1: Consider this bad boy moved.

services/ui-src/src/setupTests.js
```diff
@@ -16,3 +16,14 @@ configure({ adapter: new Adapter() });
 global.window.env = {
   API_POSTGRES_URL: "fakeurl",
 };
+
+global.matchMedia =
```
bodnarbm, line 20: Did it not work with the way you had it before, but in this file?

I think we use the same method you had before in the MCR repo (https://github.com/CMSgov/mdct-mcr/blob/ddf6d9de6f06687d849d40fec7a1a2a586b2afd5/services/ui-src/src/utils/testing/setupJest.tsx#L8-L23), so just wanted to understand what lead to the changed structure here?

bodnarbm (approved):

karla-vm (approved):

PR #2483, opened 2022-05-18
Implement Sticky Footer

ntsummers1:
# Description

Fixes [MDCT-224](https://qmacbis.atlassian.net/browse/MDCT-224). Takes the footer and makes it sticky! It'll now stay put to the bottom of the page whether theres content above it or not. I would love any feedback on any naming conventions or if you would have implemented this differently. Also it won't let me assign this to myself? Haven't run into that before.

Heres an example of where you can see it stick to the bottom:

stickyfooter

## How to test

Simply go to any page that contains the footer and you'll be able to see it in action! A specific page to look at though is the [user/profile](http://localhost:3000/user/profile) page

## Dependencies

N/A

# Get it done

Now that the PR is open this is what happens next

## Code authors checklist

- [x] I have performed a self-review of my code
- [ ] I have added thorough tests. [Testing Doc](https://qmacbis.atlassian.net/wiki/spaces/CM/pages/2914025525/Test+Suite+and+Testing+Research)
- [x] Necessary analytics were added, or no new analytics were required
- [ ] I have made corresponding changes to the documentation
- [x] I have assigned the PR to myself

## Reviewers Checklist (Two different people)

- [ ] I have done the deep review and verified the items in the above checklist are g2g
- [ ] I have done the lgtm review

## Assignee

- [ ] I have closed the PR after the review and necessary changes (squashing preferred)

PR #2775, opened 2022-05-24
MDCT 292 - Remove autosave where not needed

ntsummers1:
# Description

Fixes [MDCT-292 - Remove autosave where not needed](https://qmacbis.atlassian.net/browse/MDCT-292). The autosave in the header should no longer show unless you're on a specific report that is meant to be autosaved. It also cleans up the interactions between webpages. Before we were not using react-router-dom's Links and were using regular a tags. While this provides for a smoother transition between pages, that might not be the intended outcome here. Please let me know if the a tags were intentional or if the new Link causes unexpected behavior.

## How to test

`./dev local`

Click around the app and checkout the various pages.

## Dependencies

None

# Get it done

## Code authors checklist

- [ ] ~~I have made corresponding changes to the documentation~~

## Assignee

### DIFF
--- a/services/ui-src/src/AppRoutes.js
+++ b/services/ui-src/src/AppRoutes.js
@@ -20,12 +20,10 @@ const AppRoutes = () => {
     return <LocalLogins loginWithIDM={loginWithIDM} />;
   }
 
-  const ShowAutoSave = window.location.pathname.split("/")[2] === "sections";
-
   const VisibleHeader =
     window.location.pathname.split("/")[1] === "reports" ||
     window.location.pathname.split("/")[1] === "coming-soon" ? null : (
-      <Header currentUser={user} showAutoSave={ShowAutoSave} />
+      <Header currentUser={user} />
     );
 
   const VisibleFooter =
@@ -39,10 +37,10 @@ const AppRoutes = () => {
       className={"App " + window.location.pathname.split("/")[1]}
       data-test="component-app"
     >
-      {VisibleHeader}
       <div className="app-content">
         <Spinner />
         <Router>
+          {VisibleHeader}
           <Home user={user} role={userRole} />
           {/* These routes are available to everyone, so define them here */}
           <Route exact path="/userinfo" component={Userinfo} />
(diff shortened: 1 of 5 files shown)


BearHanded (approved):

services/ui-src/src/AppRoutes.js
```diff
@@ -36,10 +37,10 @@ const AppRoutes = () => {
       className={"App " + window.location.pathname.split("/")[1]}
       data-test="component-app"
     >
-      {VisibleHeader}
```
gmrabian, line 39: Header snuck back into the content - we still want to go by the change in [this PR](https://github.com/CMSgov/cms-carts-seds/pull/2483)?

services/ui-src/src/AppRoutes.js
```diff
@@ -36,10 +37,10 @@ const AppRoutes = () => {
...
       <div className="app-content">
         <Spinner />
         <Router>
+          {VisibleHeader}
```
gmrabian, line 43: Header snuck back into the content - we still want to go by the change in [this PR](https://github.com/CMSgov/cms-carts-seds/pull/2483)?

gmrabian:
Overall good, just that one comment

services/ui-src/src/AppRoutes.js
ntsummers1: This is intentional on my end and I double checked with Brax this morning about this to make sure that was ok. The solution I went with here uses the Router's location prop to determine whether or not the autosave should be shown. I actually went with a different approach initially where I grabbed the pathname from the url instead [See the original commit here](https://github.com/CMSgov/cms-carts-seds/pull/2775/commits/0f12c18e72e300dc4634e79f02a32c52b3d2b18a). That original approach is fine, but it was a trade off of either using react-router-doms link implementation or using straight a tags. 

There is of course another option where we do a lot of state passup where we have the components being shown determine where its shown in the header themselves but that felt like way too much work.

juliuscaesar (approved):

PR #2808, opened 2022-05-27
MDCT-292: Autosave Hotfix

ntsummers1:
# Description

Hotfix for the Autosave showing in the header. This should now make it so the autosave only shows for State users who are viewing a report that's status is Not Started, In Progress, or Uncertified. 

## How to test

`./dev local`

- Sign in as stateuser2@test.com. Only state users should ever see the autosave in the header.
- Click on a report. If it is In Progress, Uncertified, or Not Started it should show the autosave in the header. Every other scenario or page should not show the autosave in the header.
- Sign out and sign in as another user (Like adminuser@test.com). 
- If the user is an admin, they should never see the autosave on any page they go to.

## Dependencies

Please spell out any dependencies for this change -- ex -- requires an install / update / migration etc

# Get it done

## Code authors checklist

## Assignee

### DIFF
--- a/services/ui-src/src/components/layout/Header.test.js
+++ b/services/ui-src/src/components/layout/Header.test.js
@@ -53,6 +53,133 @@ const noUserNameStore = mockStore({
   },
 });
 
+const stateUserWithReportInProgressStore = mockStore({
+  stateUser: {
+    name: "Alabama",
+    abbr: "AL",
+    imageURI: "/img/states/al.svg",
+    currentUser: {
+      username: "stateuser2@test.com",
+      state: {
+        id: "AL",
+      },
+      role: "mdctcarts-state-user",
+      lastname: "States",
+      firstname: "Frank",
+      email: "stateuser2@test.com",
+    },
+    localLogin: false,
+  },
+  global: {
+    formName: "CARTS FY",
+    largeTextBoxHeight: 6,
+    isFetching: false,
+    url: "/sections/2021/00",
+    queryParams: "",
+    currentYear: 2021,
+    formYear: 2021,
+    stateName: "Alabama",
+  },
+  reportStatus: {
+    AL2021: {
+      status: "in_progress",
+      year: 2021,
+      stateCode: "AL",
+      lastChanged: "2021-01-04 18:28:18.524133+00",
+      username: "al@test.com",
+      programType: "combo",
+    },
+  },
+  save: {
+    error: false,
+    saving: false,
+  },
+});
+
+const stateUserWithReportCertifiedStore = mockStore({
+  stateUser: {
+    name: "Alabama",
+    abbr: "AL",
+    imageURI: "/img/states/al.svg",
+    currentUser: {
+      username: "stateuser2@test.com",
+      state: {
+        id: "AL",
+      },
+      role: "mdctcarts-state-user",
+      lastname: "States",
+      firstname: "Frank",
+      email: "stateuser2@test.com",
+    },
+    localLogin: false,
+  },
+  global: {
+    formName: "CARTS FY",
+    largeTextBoxHeight: 6,
+    isFetching: false,
+    url: "/sections/2021/00",
+    queryParams: "",
+    currentYear: 2021,
+    formYear: 2021,
+    stateName: "Alabama",
+  },
+  reportStatus: {
+    AL2021: {
+      status: "certified",
+      year: 2021,
+      stateCode: "AL",
+      lastChanged: "2021-01-04 18:28:18.524133+00",
+      username: "al@test.com",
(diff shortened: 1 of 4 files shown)


services/ui-src/src/components/layout/Header.test.js
```diff
@@ -53,6 +53,133 @@ const noUserNameStore = mockStore({
   },
 });
 
+const stateUserWithReportInProgressStore = mockStore({
```
ntsummers1, line 56: I'm not sure that me creating states like this is the best to do. Would it be helpful in this case to pull these mocked stores off into a different fill to reference when we want to test different state scenarios?

services/ui-src/src/store/selectors.js
```diff
@@ -157,11 +157,12 @@ export const getCurrentReportStatus = (state) => {
   if (stateUser.currentUser.role === UserRoles.STATE) {
     currentReport = `${stateUser.abbr}${global.formYear}`;
   } else {
+    if (formData?.[0] === undefined) return { status: "" };
     currentReport = `${formData[0].stateId}${formData[0].year}`;
   }
 
   const status = reportStatus[currentReport];
-  return status;
+  return status !== undefined ? status : { status: "" };
```
ntsummers1, line 165: These updates here help make this function more robust and fix crashes that could happen when accessing the state, but these fixes could be fleshed out if `{ status: "" }` is not the intended fallback

services/ui-src/src/components/layout/Header.test.js
```diff
@@ -69,9 +196,25 @@ const headerWithNoUsername = (
   </Provider>
 );
 
-const headerWithAutosave = (
-  <Provider store={store}>
-    <MemoryRouter initialEntries={["/views/sections/"]}>
+const headerOnReportPageAsStateUserThatsInProgress = (
```
ntsummers1, line 199: See above comment

services/ui-src/src/store/selectors.js
braxex: Could maybe do something like this, but again I'm not super familiar with the use case or why we'd be returning a value in one instance and an object in another. 🤷 
`return status || { status: "" }`

ntsummers1: Reason I added this was that some of the tests that are run in the header don't mock state with a reportstatus at all. When the application itself is running, the default state for reports is an empty object, {}. But when the jest tests run on, they run into this line of code here that was added with this code:
Screen Shot 2022-05-31 at 3 40 09 PM

Here the header has a new prop that grabs a reports status. So when the test runs it calls this prop and finds that report status doesnt exsist or if it does exist as an empty object then it tries to illegally access the formData field (See line 160 in selectors.js for what I'm talking about). This code takes this error possibilities into account. 

Not to say that maybe this code overreaches and we could simply just update the mocked stores to include the reports... This was my thought process though.

services/ui-src/src/store/selectors.js
```diff
@@ -147,7 +147,7 @@ export const selectSectionsForNav = (state) => {
  * @returns {object} The reportStatus object associated with the current report
  */
 export const getCurrentReportStatus = (state) => {
-  if (state.reportStatus.status === null) {
+  if (!state.reportStatus || state.reportStatus.status === null) {
```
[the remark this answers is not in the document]
ntsummers1: Expanded thought process behind this code here: https://github.com/CMSgov/cms-carts-seds/pull/2808/files#r886070711

services/ui-src/src/components/layout/Header.test.js
ntsummers1: Pulled out into [services/ui-src/src/store/fakeStoreExamples.js](https://github.com/CMSgov/cms-carts-seds/pull/2808/files#diff-20a7089ac9666d9a1f177c99b103c8be53e5d84812cc92f7edc63a36d5112617)

gmrabian (dismissed):

services/ui-src/src/store/selectors.js
braxex: Yeah I think we should update the mocked stores and anywhere that's relying on there being a status to be able to handle the report state being `{}` instead of sometimes returning `status` and sometimes returning `{ status: "" }`. 'Cause ideally a return statement should not be returning different types.

services/ui-src/src/store/selectors.js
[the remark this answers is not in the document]
braxex: Left a comment on the other chain about maybe trying to update things here a bit. Let me know if you want to pair on this at some point.

braxex (approved):

BearHanded (approved):

PR #2861, opened 2022-06-08
MDCT-270 - Carts V3 -> Prince

BearHanded:
# Description
* One of the 2 PRs for MDCT-270. Addresses the ability to call to prince with the print component.
* Also allows state users to post to non state path param endpoints, so they can use printPdf. Added protection for the state being required to the one other file affected.

## How to test
Checkout this branch!
https://d1corn20m4y9yl.cloudfront.net/

You should be able to print!

Use incognito or guru will break it

## Dependencies
Either use the above link for testing in AWS, or use credentials from cloudtamer -> Carts Dev
If you opt for local, you also need the host & path for prince (yes they're separate for signing purposes, I don't like arbitrary string splitting)

# Get it done
Free me from this AWS config nightmare

## Code authors checklist

- [ ] I have assigned the PR to myself

## Assignee


braxex (dismissed):
Looks good 👍

services/app-api/serverless.yml
```diff
@@ -82,8 +84,11 @@ provider:
             - dynamodb:PutItem
             - dynamodb:UpdateItem
             - dynamodb:DeleteItem
+            - execute-api:Invoke
```
gmrabian, line 87: Should put this permission in its own list item. Usually for AWS resources you make a separate item per resource type.
```
- Effect: "Allow"
   Action:
     - execute-api:Invoke
   Resource: "*"
```

services/ui-src/src/components/sections/Print.js
```diff
@@ -6,11 +6,13 @@ import { Button } from "@cmsgov/design-system";
 import PropTypes from "prop-types";
 import Title from "../layout/Title";
 import Section from "../layout/Section";
-import axios from "../../authenticatedAxios";
+//import axios from "../../authenticatedAxios";
```
gmrabian, line 9: still need this?

services/app-api/libs/authorization.ts
```diff
@@ -44,13 +44,8 @@ export const isAuthorized = (event: APIGatewayProxyEvent) => {
...
   // if user is a state user - check they are requesting a resource from their state
-  if (userRole === UserRoles.STATE) {
-    if (userState && requestState) {
-      return userState.toLowerCase() === requestState.toLowerCase();
-    }
-    if (!requestState && event.httpMethod != RequestMethods.GET) {
-      return false;
-    }
+  if (userRole === UserRoles.STATE && userState && requestState) {
+    return userState.toLowerCase() === requestState.toLowerCase();
```
gmrabian, line 48: I think you lose a case here where if requestState doesn't exist you should return false. In this case it would return true. Maybe you figured out that case isn't needed though.

gmrabian:
Happy with this. Just small comments.

services/app-api/serverless.yml
BearHanded: will do, thanks for the feedback

services/app-api/libs/authorization.ts
BearHanded: We are scrapping that case. We have need of API endpoints without that path param and accessible by State Users. The one non-get request that was relying on this behavior has been shored up to require a state in the path explicitly.

gmrabian (approved):

braxex (approved):

PR #3596, opened 2022-08-08
[Mdct-443] CHIPS enrollments lookups

BearHanded:
# Description

[Ticket](https://qmacbis.atlassian.net/browse/MDCT-297)
Add lookup functionality for the SEDS CHIPS enrollment counts.
Also includes the fix for the missing logout menu. One liner in which a bad prop was being referenced.

**Previous Behavior:**
* CARTS V2 waited on messages from SEDS -> Kafka, and wrote entries into the stg_enrollment_counts table
* A trigger on that table then wrote values directly into the carts sections

**New Behavior**
* Aggregate messages into the new enrollment counts table (kafka integration yet to come)
* Forms can now reference a lookupEnrollmentCounts field property, that will load data for that field when viewing the form in the same way that ACS & FMAP data is loaded.
* 2022 forms onward can use this property. 2021 forms already have the values set, so don't need to change.
* Additionally migrate historical enrollment counts into the v3 database, just to track that data in case anything in the future needs to reference it.
* hideSections functionality needed an update to be able to reference this data. It previously only handled single answers, or the very first entry in a table when the data was computed. The update now should be able to handle computed enrollment counts, computed tables, and computed tables in which some data is computed, and others are not.
* New API endpoint that is called when loading a form to get that year's enrollment counts if they exist. Also called when rendering the Print page.

## How to test

Locally: 
* Update the env with new settings in the example env for the new table.
* As an admin user generate 2022 forms (The seeded 2022 form uses the new lookup format)
* As a state or admin user, section 2 should display the seeded CHIPs data in question one, and because the % is greater than 3, it should display the following conditional question.
* If you're checking it out as admin, AL should have data populating, and  CT should not have data available

On a branch, you would need to modify a states form to use the new format, but the data should be imported to support it.

## Dependencies
New env vars.

# Get it done

Review it

## Code authors checklist

## Assignee


gmrabian (approved):
Looks good locally

karla-vm (approved):
Walked through it with @BearHanded and everything looks good ✨

PR #139363, opened 2023-05-25
Carts section 5 fix

britt-mo:
### Description
Section 5 had incorrectly labeled headers and labels for `lookupFmapFy` labels. That was changed but this is a data fix for any programs that were already filled with the incorrect labels.

### Related ticket(s)

MDCT-2320

---
### Important updates

---
### Author checklist

- [ ] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
---


BearHanded (approved):
I think this looks good. We could definitely format it a bit better, but since we'll run it and archive it immediately it's less pressing

PR #139446, opened 2023-09-07
Add some tests for shouldDisplay()

benmartin-coforma:
### Description
This PR contains no code changes, only new tests.

The `shouldDisplay()` function is used to determine whether certain parts of the app should... display. Depending on the current user, current program, and various other pieces of application state. Prior to this PR, it was one of the largest untested chunks of code in the app. In fact the majority is _still_ untested - we have only 40.42% line coverage on this file. But we that is up from 6.38%, so I call it a win.

The portion of this file that remains untested relates to `hide_if_table_value`. Its behavior is tremendously complicated, despite being used by only 2 questions per year, which use only a fraction of that behavior. It would be better for us to delete as much of that code as possible, rather than write new tests for it.

This PR increases CARTS' overall unit test coverage for `ui-src` from 45.03% to 47.03% by line count. ✨ +2% ✨ 

### Related ticket(s)
MDCT-2807

---
### How to test
```bash
cd services/ui-src
yarn test --coverage
```

### Important updates
n/a

---
### Author checklist

- [x] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary

### DIFF
--- a/services/ui-src/src/util/shouldDisplay.test.js
+++ b/services/ui-src/src/util/shouldDisplay.test.js
@@ -1,0 +1,314 @@
+import { shouldDisplay } from "./shouldDisplay";
+import { AppRoles } from "../types";
+
+describe("shouldDisplay", () => {
+  it("should display everything for CMS Admins", () => {
+    const state = {
+      stateUser: {
+        currentUser: {
+          role: AppRoles.CMS_ADMIN,
+        },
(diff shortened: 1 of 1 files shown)


gmrabian (approved):
neato

BearHanded (approved):

PR #139460, opened 2023-09-19
Enable field 2 in Basic State Information

gmrabian:
### Description
Make Question 2, Program Type, in Basic State Information editable, and update corresponding logic to make form react to this update

Three tables hold information for `programType`:
- state
- sections
- state-status

state – holds a static value and thus we should seek to neither update nor read `programType` from this in the app's flow
sections – no change, holds the `programType` value from the user's choice in the UI
state-status – is now updated with the value from sections, and is the sole reference for that value in the frontend because it is accessible to all user types

### Related ticket(s)
MDCT-2993

---
### How to test

- Login as admin
- Generate 2022 and 2023 forms
- Login as stateuser2 and select the 2023 form
- Question two of Basic State Information should be editable
- The blue banners in Section 1 should react to the changes (Part 1 and 3 get bannered when you select the third option; parts 2 and 4 when you choose the second; none bannered for the first). There are other pages in the form that also change as a result and they can be identified by searching `show_if_state_program_type_in` in seed-section-base-2023.json

---
### Author checklist

- [x] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
---


services/app-api/handlers/formTemplates/post.ts
```diff
@@ -60,7 +60,6 @@ export const post = handler(async (event, _context) => {
     stateStatuses.push({
       stateId: state.code,
       year: yearNumber,
-      programType: state.programType,
```
gmrabian, line 63: more investigation needed on this needed but for now I don't see a strong case for programType being in state statuses. If we want to keep it we will have to include programType that in the update call.

gmrabian (commented):
Initial refactor. There are other places that use programType differently and will need to be remedied. This is just a start.

services/app-api/handlers/formTemplates/post.ts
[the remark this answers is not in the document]
BearHanded: I am leaning towards keeping it and updating it, mostly because it is something that is currently being fed to kafka and it keeps any downstream consumption the same.

services/ui-src/src/store/formData.js
```diff
@@ -34,13 +34,6 @@ export default (state = initialState, action) => {
           updatedData[0].contents.section.subsections[0].parts[0].questions[0] =
             lastYearData[0].contents.section.subsections[0].parts[0].questions[0]; // Name
         }
-        if (
-          !updatedData[0].contents.section.subsections[0].parts[0].questions[1]
-            .answer.entry
-        ) {
-          updatedData[0].contents.section.subsections[0].parts[0].questions[1] =
-            lastYearData[0].contents.section.subsections[0].parts[0].questions[1]; // Type
-        }
```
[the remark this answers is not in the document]
BearHanded: If it is already pre-set, this is fine, as long as we show it works.

services/ui-src/src/util/shouldDisplay.js
```diff
@@ -175,19 +176,10 @@ const hideIfTableValue = (state, hideIfTableValueInfo) => {
...
-      reportStatusCode = state.stateUser.abbr + state.global.formYear;
-    } else {
-      reportStatusCode = state.formData[0].stateId + state.formData[0].year;
-    }
-    program = state.reportStatus[reportStatusCode].programType;
-  } else {
-    program = programType;
-  }
+  program =
+    programType ?? selectFragmentById(state, "2023-00-a-01-02").answer.entry;
```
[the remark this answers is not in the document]
BearHanded: this works across sections right? Haven't pulled it down to test.

This is also explicitly calling out the year hear which won't work on older forms

services/ui-src/src/store/formData.js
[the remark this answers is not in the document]
BearHanded: Does this also mean things get weird if we make this change and items have been previously mis-aligned?

services/ui-src/src/util/shouldDisplay.js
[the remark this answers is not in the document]
gmrabian: I fixed this to read year from the formYear and verified it reads correctly across previous years. Also ran through all sections in 2023 and they rendered with the conditionals as expected (did not check previous years yet)

services/app-api/handlers/section/update.ts
```diff
@@ -89,10 +90,37 @@ export const updateSections = handler(async (event, _context) => {
...
+        programType: getProgramType(year, reportData, stateStatus?.programType),
       },
       "post"
     ),
   };
 
   await dynamoDb.update(statusParams);
 });
+
+const PROGRAM_TYPE_QUESTION_ID = "-00-a-01-02";
```
gmrabian, line 102: this is the only place this is used so I left it here rather than in a constants file

services/app-api/handlers/section/update.ts
```diff
@@ -89,10 +90,37 @@ export const updateSections = handler(async (event, _context) => {
...
+    return currentYearProgramType;
+  }
+  // attempt to find programType from the previous year's form
+  const previousYear = parseInt(year) - 1;
+  idExpression = `${previousYear}${PROGRAM_TYPE_QUESTION_ID}`;
+  jpexpr = `$..*[?(@ && @.id=='${idExpression}')]`;
+  fragment = JSONPath({ path: jpexpr, json: reportData[0].contents });
+  const previousYearProgramType = fragment?.[0]?.answer?.entry;
+  return previousYearProgramType ?? currentProgramType;
+};
```
gmrabian, line 126: checks current and previous year for updated programType from form, otherwise uses existing programType from status. I simplified the frontend to only read programType from status so this keeps everything in sync behind the scenes

services/ui-src/src/components/layout/Part.js
```diff
@@ -100,49 +87,15 @@ const mapStateToProps = (state, { partId }) => {
...
-  const userState = location[3]; // Current state, ie: "AL" or "CT"
-  const programData = state.allStatesData.find(
-    (element) => element.code === userState
-  );
 
   return {
     context_data: part.context_data,
     questions,
-    show: showPartBasedOnUserType(contextData, programData, state),
+    show: shouldDisplay(state, contextData),
```
gmrabian, line 94: programType is no longer needed as an argument for shouldDisplay so we can eliminate two functions from this file and really simplify things

services/ui-src/src/store/stateUser.js
```diff
@@ -16,7 +16,6 @@ export const getUserData = (userObject) => {
 
 export const getProgramData = (programObject) => ({
   type: PROGRAM_INFO,
-  programType: programObject.programType,
```
gmrabian, line 19: we don't want programType available in state information because that table has static values for programType

services/ui-src/src/util/shouldDisplay.js
```diff
@@ -201,6 +188,8 @@ const shouldDisplay = (state, context, programType = null) => {
    * displaying relies on that answer being included in the show_if_state_program_type_in array
    */
   if (context.show_if_state_program_type_in) {
+    const reportStatusCode = state.formData[0].stateId + state.formData[0].year;
+    const program = state.reportStatus[reportStatusCode].programType;
```
gmrabian, line 192: greatly simplifies logic in this function. all users can access programType from report status now so that is the only place we need look for it

gmrabian (commented):
Check it out!

BearHanded (dismissed):
Looked at this pretty thoroughly, and this looks solid, but I haven't had a chance to run it locally. Will update if I do.

benmartin-coforma (dismissed):
Very nice. Especially like that you updated the shouldDisplay tests for clarity 👍

BearHanded (changes requested):
Ran this locally and tested it against rendering sections. I am seeing that the browser needs to refresh before rendering the new state for a section.

Replication steps:
- Start with the report as a "combo"
- Switch to Medicaid Expansion CHIP Only
- Navigate to 3F - The form appears when it shouldn't
- Refresh, you should now see the correct state

BearHanded (approved):

benmartin-coforma (approved):

PR #139486, opened 2023-10-12
Cypress fill test

benmartin-coforma:
### Description
This PR adds one (1) integration test to CARTS, which will interact with some (some) of the fields in Section 1.

### Related ticket(s)
CMDCT-2806

---
### How to test
Observe that the E2E tests on this branch include the new test, and that it passed.

### Important updates
n/a

---
### Author checklist

- ~[ ] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary~

### DIFF
--- a/services/ui-auth/libs/users.json
+++ b/services/ui-auth/libs/users.json
@@ -402,7 +402,7 @@
       },
       {
           "Name": "custom:cms_state",
-          "Value": "DC"
+          "Value": "AL"
       }
     ]
   },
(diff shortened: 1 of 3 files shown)


gmrabian (commented):
I am curious why you modified the users and cypress config 🤔 

E2E tests pass. Code climate check doesn't but the file it calls out isn't included in this PR. Maybe just pull in main if they have diverged? Idk

Will approve once we know everything is ok here

BearHanded (approved):

gmrabian (approved):
discussed with Ben separately: the state assigned to the user will not update for already provisioned users, so the original change did not take effect. This is aligning the expected state with actual state.

PR #139488, opened 2023-10-13
Remove styled-components package

benmartin-coforma:
### Description
This PR removes `styled-components` from CARTS.

`styled-components` has a reported "vulnerability": when installing certain versions, if it believes you are in Russia, it will print out a protest message regarding Ukraine. I don't personally see a problem with this, but our security scans do.

As it happens, we don't actually use `styled-components` within CARTS; it is installed and then ignored. So I'm taking this as an opportunity to dismount the dependency upgrade treadmill, for this package at least.

### Related ticket(s)
[CMDCT-2855](https://jiraent.cms.gov/browse/CMDCT-2855)

---
### How to test
* Observe that the app builds without error
* Observe that all tests pass
* Click around in the app, and observe that everything seems to render fine

### Important updates
N/A

---
### Author checklist

- ~[ ] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary~

### DIFF
--- a/services/ui-src/package.json
+++ b/services/ui-src/package.json
@@ -47,7 +47,6 @@
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.3.0",
     "sass": "^1.62.0",
-    "styled-components": "^5.2.1",
     "underscore": "^1.11.0"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


gmrabian (dismissed):
Didn't see any degradation
Also
✊ 🇺🇦 🌻

BearHanded (dismissed):

benmartin-coforma:
I had to merge main & re-run `yarn` in ui-src. I'm setting this to auto-merge once it gets another approval, because it already had two.

gmrabian (approved):

PR #139504, opened 2023-10-30
Add tests for UploadComponent

benmartin-coforma:
### Description
This PR adds tests to `UploadComponent.js`. It was our most undertested file, as measured in lines of code.

This bumps CARTS/services/ui-src line coverage by 4.5%, up to 57.34.

There are a few "TODO" type items in there at the moment; some cleanup and maybe an extra test or two would be appropriate before merging.

My goal was to _only_ increase coverage, with _zero_ code changes. I failed this goal, because I found two clear bugs with clear apparent fixes:
1. The XMLHttpRequest to upload a file to S3 is sent before a listener is attached for its return. I re-ordered those statements.
2. The ability to upload multiple files at once is broken by a `return` within a `for` loop. I removed the return.

Additionally, I added some `data-testid` attributes for ease of locating elements. It would be better if we could find a semantic way to locate these elements, but I don't want to rearrange the component without a story either.

_EDIT: Having achieved a high level of coverage with very "deep" tests, I became much more comfortable doing an in-place refactor. I pulled all of the direct API code out of UploadComponent and stashed it in a new file, `fileApi.js`._

### Related ticket(s)
n/a

---
### How to test
```bash
cd services/ui-src
yarn
yarn test --coverage
```

### Important updates
n/a

---
### Author checklist


benmartin-coforma:
> The test coverage on the diff in this pull request is **66.6%** (**90%** is the threshold).
> 
> This pull request will bring the total coverage in the repository to **65.4%** (**3.5%** change).

I do not know how CodeClimate is calculating the test coverage on the diff here, but it is definitely higher than 66%. All but one of the seven changed lines in `UploadComponent.js` is exercised by unit tests, and the number of uncovered lines in that file is down from 214 to 3.

The untested behaviors of UploadComponent include:
* When the user says they don't want uploaded files displayed, and upload files while that preference is active, the uploaded files are not added to the display.
* Admin users cannot upload files.
* Files cannot be uploaded to a report that is not in progress (for example, when it is certified).
* When uploading to S3 results in an error, the upload component displays that error gracefully. And for what it's worth, I did write this test, but "gracefully" was never defined and would require nontrivial code changes to implement.

Coverage summary for ui-src prior to this change:
```
Statements   : 53.41% ( 884/1655 )
Branches     : 39.11% ( 336/859 )
Functions    : 44.93% ( 173/385 )
Lines        : 52.88% ( 843/1594 )
```

And after this change:
```
Statements   : 57.7% ( 955/1655 )
Branches     : 41.32% ( 355/859 )
Functions    : 49.61% ( 191/385 )
Lines        : 57.34% ( 914/1594 )
```

In short, the numbers don't add up.

gmrabian:
My guess as to why the coverage is so low is because of apiLib. It went from an F->D, which is an improvement, but it's probably averaging that against `UploadComponent` and ending up with 66 (not exactly sure how). `apiLib` isn't touched here, so it could be that it's now getting tested and it wasn't before? Is there a place we should be mocking where we are not?

Screenshot 2023-10-30 at 1 45 07 PM

gmrabian (commented):
I like how much coverage this gets! I feel like the mocks are very deep in the process. If it works it works, but could we clean that up by getting closer to the surface?
i.e. mock call returns 200 vs. checking for http verb, url, headers, etc.
Maybe I'm not understanding the necessity here. Feel free to ping me and fill me in. Thanks!

gmrabian (approved):

britt-mo (approved):
Super thorough! 🤯

PR #139507, opened 2023-11-06
Add unit tests for the Date Range component

benmartin-coforma:
### Description
This PR adds a bunch of tests around `<DateRange>`.

NOTE: There are four changes to the actual component - I added labels to the four inputs. As far as I can tell this does not affect the display. It definitely improved the testability, because I use those labels to locate the inputs on the screen. And I believe it is good for accessibility? Would love feedback specifically on this.

### Related ticket(s)
n/a

---
### How to test
* Take a look at a date range in the UI. There is one in Section 2, which appears if you answer Yes to question 3 ("Do you have any alternate data sources...")
* Run the unit tests

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui-src/src/components/layout/DateRange.js
+++ b/services/ui-src/src/components/layout/DateRange.js
@@ -246,7 +246,7 @@ class DateRange extends Component {
               data-test="component-daterange-monthstart"
               name="monthStart"
               numeric
-              label=""
+              label="range start month"
               onChange={this.handleInput}
               onBlur={this.validateStartInput}
               value={monthStart}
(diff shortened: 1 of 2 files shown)


gmrabian (approved):

britt-mo (approved):
I agree with your comment in the test file, I think the error message should only appear if the input field has been touched already (as its own card)
Screenshot 2023-11-06 at 11 57 01 AM

Also I'm not entirely sure about the `label` field from an accessibility standpoint. Can't hurt to add it but I also didn't see the label in the `elements` tab which leads me to believe it isn't actually being used here beyond your testing purposes.

PR #139519, opened 2023-11-16
Change incorrect money types to integer - etl and seed update

gmrabian:
### Description

Six fields in Section 5 incorrectly have type "money" when they should be "integer". This updates the templates and seed data (for local dev) and contains an ETL to run in each environment to fix existing reports.

### ETL
- Looks for these six questions and updates `type` attribute to `integer`
Note: I checked if there are differences between  money and integer in validation. All money values must be whole numbers, and all periods, commas, dollar signs are removed upon type/blur/save, so there should be no issue with changing the type if users have already entered values. 

ETL only runs for 2023. DK said leave 2022 as is.
The lambda for the ETL can be safely removed after this change is in every environment | 

### Seed updates: 
- update 2022 and 2023 seed templates, just so when we make future reports (2024+) we have a more accurate base
  - also means when we generate 2023 locally and in deployed branches it has correct type
- update 2022 seed-section to have correct type, for local dev only

Note: I verified that the question numbers `2023-05-a-03-01*` and `2023-05-a-04-01*` are only used to display in the table below those questions, not anywhere else for calculations or display.

### Related ticket(s)

CMDCT-3048

---
### How to test

Test seed files updated
- Run CARTS
- Generate 2023 forms as admin
- As state user, go to 2023 form, section 5, parts 3 and 4
- Ensure the questions about "How many children were eligible for managed care..." in each part correctly shows an integer field, not a money field (no dollar sign)

Test ETL script
- Login to the CARTS-dev AWS account
- Open dynamodb and lambda in two tabs
  - Dynamo table: `cmdct-3048-section`
  - Lambda: `database-cmdct-3048-convertMoneyTypeToInteger`
- In another tab go to the deployed ui: `https://de853vni0o7zv.cloudfront.net`
- Login as stateuser2 and go to `/sections/2023/05`
- Scroll down to Parts 3 and 4. See if the first question box begins with a dollar sign or not.
- If not
  - Find `AL-2023` in the dynamo table and click on the section 5 entry
  - Go to JSON view
  - Search for `2023-05-a-03-01-` and change the types of `a,b,c` to "money" (resetting to verify the fix)
  - Repeat for `2023-05-a-04-01-`
- On the lambda, run "test" in the test tab. It should complete in seconds.
- Refresh the UI and see that the dollar signs are gone, and any number value remains the same

See demo video:

https://github.com/Enterprise-CMCS/macpro-mdct-carts/assets/57802560/2e11cf70-0d0f-4ee0-a0bf-95623fc5bd56

---
### Author checklist

---

### DIFF
--- a/services/database/data/seed-local/seed-section.json
+++ b/services/database/data/seed-local/seed-section.json
@@ -11665,7 +11665,7 @@
                             "comment": "This is the first real question so far in this part…",
                             "id": "2022-05-a-03-01-a",
                             "label": "2022",
-                            "type": "money",
+                            "type": "integer",
                             "answer": {
                               "entry": "5"
                             }
@@ -11673,7 +11673,7 @@
                           {
                             "id": "2022-05-a-03-01-b",
                             "label": "2023",
-                            "type": "money",
+                            "type": "integer",
                             "answer": {
                               "entry": "5"
                             }
@@ -11681,7 +11681,7 @@
                           {
                             "id": "2022-05-a-03-01-c",
                             "label": "2024",
-                            "type": "money",
+                            "type": "integer",
                             "answer": {
                               "entry": "5"
                             }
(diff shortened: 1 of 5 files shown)


BearHanded (approved):

benmartin-coforma (approved):
I haven't tested, but the code looks good.

PR #139559, opened 2024-02-08
Fix infrastructure destroy action

berryd:
### Description

The purpose of this pull request is to correct destroy actions for Snyk or Dependabot created branches by providing a consistent means to generate shortened stage names from autogenerated branches that may exceed Cloudformation length requirements when prefixed to services managed by Cloudformation stacks.  Additionally, this pull request adds the ability to trigger destroy actions manually against any non-protected target using any Github hosted branch.

- `GITHUB_REF` environment variable is the same as `github.ref` and includes the fully qualified path
- Destroy actions always trigger on the default branch so `GITHUB_REF`/`github.ref` would be `/refs/heads/master`
- Destroy actions know what the target is based on `github.event.ref` which is NOT the fully qualified path
- Defining `workflow_dispatch` enables manual triggering destroy actions against a given code base and enables specifying target.

The changes in this pull request do the following:
- Apply consistent lookups in Deploy (concurrency lock would be deploy-[branch] rather than deploy-/refs/heads/[branch])
- Remove hard coded addition to destroy action which would break if a tag is deleted
- Enable manual dispatching of events.

For other repositories which may differ in structure, it is suggested that the `workflow_dispatch` be added as an initial pull request, as this has to make it to the default stage before it can be acted upon.  This will enable development against destroy actions which can only trigger from the default branch until the configuration change is merged.

### Related ticket(s)

CMDCT-3256

---
### How to test

Once this has been merged...

Create three branches

dependabot/npm_and_yarn/berry-did-this
snyk-upgrade-berry-did-this9faf06c882b46cc4a8e83
berry-test
Push and wait for builds to succeed.

Delete the remote branches and verify the destroy step completes.

Create one more branch, any arbitrary name, and add echo this is a test && exit 1 to the destroy yaml file at line 37. Exercise the manual trigger and verify the update in the unmerged branch is executed in the destroy step

### Important updates

No functional changes should be experienced during normal development activities.  The changes do include some minor refactoring of how references are leveraged, but the functional changes should apply to Snyk and Dependabot created branches with the addition of Github actions manual triggering from the destroy workflow.

As of the time of writing of this pull request, the following are the remote branches for this product, and the highlighted ones are the ones this pull requests asserts to be protected:
  origin/bigmac-sdk-v3
  origin/deployed-login
  origin/docraptor-poc-3249
  origin/entjira
  origin/fix-snykdependa-destroy
  origin/iam
  **origin/main** <
  origin/master
  **origin/production** <
  origin/snyk-fix-99615638953ff7b1ad7a664beb9a511f
  origin/snyk-upgrade-3e2e25b088f8510923cf3077dfec294f
  origin/snyk-upgrade-448443e6ea2d600363dda07c986b9de0
  origin/snyk-upgrade-783178dddacd564062a91dc354304dc0
  origin/tealium-3288
  **origin/val** <

---
### Author checklist

---

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,12 +23,11 @@ jobs:
     env:
       SLS_DEPRECATION_DISABLE: "*" # Turn off deprecation warnings in the pipeline
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: set branch_name # Some integrations (Dependabot & Snyk) build very long branch names. This is a switch to make long branch names shorter.
         run: |
           BRANCH_NAME=$(./.github/setBranchName.sh ${{ github.ref_name }})
           echo "branch_name=${BRANCH_NAME}" >> $GITHUB_ENV
-      - uses: actions/checkout@v3
       - name: Validate branch name
         run: ./.github/branchNameValidation.sh $STAGE_PREFIX$branch_name
       - name: set branch specific variable names
@@ -103,7 +102,7 @@ jobs:
     if: ${{ github.ref_name != 'production' }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Run Cypress Tests
         uses: cypress-io/github-action@v5
         with:
@@ -133,7 +132,7 @@ jobs:
     if: ${{ github.ref_name != 'production' }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Check Project A11y
         uses: cypress-io/github-action@v5
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -23,16 +23,11 @@ jobs:
     env:
       SLS_DEPRECATION_DISABLE: "*" # Turn off deprecation warnings in the pipeline
     steps:
+      - uses: actions/checkout@v3
```
gmrabian, line 26: checkout@v4 is out if we want to upgrade while we're here 🤷 not really relevant though

gmrabian (dismissed):
Looks good

CARTS also has a `master` branch from before we migrated to `main` and the new infrastructure set. I'm pretty sure it's at least already partially deleted, and we don't care if it gets fully deleted. just calling it out

BearHanded (approved):

PR #139569, opened 2024-02-20
Text fix script – Dental Code Typo

gmrabian:
### Description

HELLO! 

This script fixes a typo in the Dental Codes
| | |
|-----|-----|
| OLD | `D0100-D9999` |
| NEW | `D1000-D1999` |

- Updated all instances in the seed JSON files (and unified `  -  ` (space hyphen space) and `–` (endash) to use endash for all of them
- Added script utils file for dynamodb to make future script fixes like this easier (and in aws-sdk-v3)
- Added script for this specific text fix, updates two question hints with new text (only change is the codes)

### Related ticket(s)

CMDCT-3203

---
### How to test

- Create a branch off CARTS main and add _only the script files_ (the seed files are correct, so local seeds will reflect the correct text if you start from this branch)
- In one terminal tab start CARTS locally
- In a new terminal tab, from CARTS root, run:
  `DYNAMODB_URL="http://localhost:8000" dynamoPrefix="local" node services/database/scripts/dental-code-text-fix.js`

Alternatively
- Create your own test branch or choose an existing one based on CARTS main
- Go to the deployed instance of that branch
  - Generate 2023 forms as an admin
  - Go to Section 3G in any 2023 report
  - Verify the third and fourth questions (3a and 4a if you answer 'No' to question 1) have D0100-D9999 in the hint text
- Get AWS credentials from carts-dev in your terminal
- From local you can run `dynamoPrefix="YOUR BRANCH NAME" node services/database/scripts/dental-code-text-fix.js`
- This will modify section 3 of that branch
- Refresh your browser and verify the associated question hints now mention D1000-D1999.
- Delete your test branch to release those resources

---
### Author checklist

---

### Looking for feedback
1. Do we want to add tests for the utils file? I think that would make sense if it's a more permanent fixture in our workflow. The tests would be very similar to the dynamodb-lib tests in app-api, which makes me wonder if there is a way to save on code duplication (monorepo and whatnot)
2. Do we want to delete the existing scripts that use AWS SDK V2 so that it's easier for future devs to pick up on the new pattern (V3, utils file, etc.)? I also think documentation could help with that.

### DIFF
--- a/services/database/README.md
+++ b/services/database/README.md
@@ -1,1 +1,19 @@
 # database
+
+## scripts
+
+The [scripts/](./scripts/) folder is for small text and data fixes. Write and test new scripts locally and with ephemeral branches. After merging, run new scripts against the AWS main, val, and production instances (preferably with a second developer for checks along the way).
+
+Any utilities shareable across scripts can be added to the [scripts/utils/](./scripts/utils/) folder. Please use AWS SDK V3, or application latest.
+
+### Running against a local instance:
+
+- In one terminal tab start CARTS locally
+- In a new terminal tab, from CARTS root, run:
+  `DYNAMODB_URL="http://localhost:8000" dynamoPrefix="local" node services/database/scripts/{script_name}.js`
+
+### Running against a deployed instance:
+
+- Get AWS credentials from the CARTS AWS account in your terminal (carts-dev for all ephemeral branches and main)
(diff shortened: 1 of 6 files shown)


BearHanded (approved):
tested locally and verified

karla-vm (approved):

PR #139591, opened 2024-03-20
Purify pdfs

benmartin-coforma:
### Description
This PR applies DOMPurify's sanitization to our HTML before sending it to DocRaptor for rendering to PDF.

Since this HTML comes from the browser and goes through our API, in theory an attacker could send their own HTML to our API, with malicious code to access the filesystem. While this does not pose a risk to us (it's DocRaptor who is evaluating the HTML, and they have their own filesystem protections in place), it's still a good idea.

This PR also contains a frontend change. If DOMPurify finds a style tag that contains CSS which might be usable for some kind of attack, DOMPurify will remove that entire style tag. We had just a few lines (of SVG code embedded in CSS) that triggered DOMPurify in this way, and I've changed them to appear safer (by running the SVG through `encodeUriComponent`).

**QUESTION FOR REVIEWERS**: How might we test against regressions here (such as new SVGs embedded in CSS in the future, which may render fine in the browser but break the PDF call)? Would it be worthwhile to do so?

### Related ticket(s)
CMDCT-3370

---
### How to test
1. Log into Carts
2. Enter a report
3. Click the print button at the bottom
4. Click the print button at the top
5. You should see a "nice" PDF. If it fails to load at all, or if the styles are out of whack (for example, tiny SVG icons expanding to take up the entire page), then you've found a bug.

### Important updates
n/a

---
### Author checklist

---

### DIFF
--- a/services/app-api/handlers/printing/printPdf.ts
+++ b/services/app-api/handlers/printing/printPdf.ts
@@ -1,6 +1,11 @@
 import handler from "../../libs/handler-lib";
 import * as logger from "../../libs/debug-lib";
+import createDOMPurify from "dompurify";
+import { JSDOM } from "jsdom";
 import { fetch } from "cross-fetch"; // TODO delete this line and uninstall this package, once CARTS is running on Nodejs 18+
+
+const windowEmulator: any = new JSDOM("").window;
+const DOMPurify = createDOMPurify(windowEmulator);
 
 /**
  * Generates 508 compliant PDF using an external Prince-based service for a given HTML block.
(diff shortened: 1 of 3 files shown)


BearHanded (approved):
re: How might we test against regressions here (such as new SVGs embedded in CSS in the future, which may render fine in the browser but break the PDF call)? Would it be worthwhile to do so?

I'm not sure. I don't of a good way to do test for a silent failure like that. I don't know if we could save the generated artifact to a release via unit tests for spot checking, but that also feels like overkill.

gmrabian (commented):
~~do we need to add `docraptorApiKey` to .env_example? My api build failed without it.~~

nevermind I see it now

gmrabian (approved):
code looks good. I don't have a key to test it yet

gmrabian (approved):
tested and looks good. some of the pdf is funky but pretty sure that's unrelated

Screenshot 2024-03-27 at 2 54 37 PM

PR #139616, opened 2024-04-23
CMDCT-3465 - Amplify V6

BearHanded:
### Description

Upgrade Amplify to V6. Also closes a vuln or two for snyk.

Amplify V6 Guide [Here](https://docs.amplify.aws/angular/build-a-backend/troubleshooting/migrate-from-javascript-v5-to-v6/).

**After this merges, we'll want to make sure the signin with real credentials works in main.**

Notable Steps for the same upgrade in future apps -

General:
- Install aws-amplify ^6.0.29 (there was a fix around 6.0.15 for error codes)
- Note the new format for the config in index.js. Rest API changed and the Auth gets nested under cognito with a number of name changes (userPoolClientId, scopes).
- update imports to be scoped to the category, unpack calls like signIn
- Update mocks

Auth:
- Sign In takes an object rather than multiple args
- Function name changes for Auth.currentAuthenticatedUser -> getCurrentUser, federatedSignin, currentSession, etc

API:
- update apiLib functions to async and await
- Read the stream in API lib as text first to test length, then convert to json, rather than using the handy .json() function, or everything explodes on an empty response.
- Added a convenience mock for amplify calls, that deals with the pair of awaits
- DELETE no longer (currently) accepts a body in the amplify library. Move any delete calls to use path params exclusively.

### Related ticket(s)

CMDCT-3465

---
### How to test

Deployed: https://d16aqmuh3yvyso.cloudfront.net/
Go up, click around.
- Get requests happen all over the place
- Posts happen all over
- Puts happen on autosave in the form
- Delete can be invoked on an uploaded file (see section 2)
- Get PDF still works (it's an api call and all)

### Notes

run `yarn`

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/hooks/authHooks/requestOptions.js
+++ b/services/ui-src/src/hooks/authHooks/requestOptions.js
@@ -2,9 +2,12 @@ import { fetchAuthSession } from "aws-amplify/auth";
 
 async function requestOptions(body = null) {
   try {
-    const { idToken } = (await fetchAuthSession()).tokens ?? {};
+    const apiKey = (await fetchAuthSession()).tokens?.idToken?.toString();
+    if (apiKey === undefined) {
+      throw new Error("Missing API Key from auth session.");
+    }
     const options = {
-      headers: { "x-api-key": idToken.toString() },
+      headers: { "x-api-key": apiKey },
     };
     if (body) {
       options["body"] = body;
--- a/services/ui-src/src/hooks/authHooks/requestOptions.js
+++ b/services/ui-src/src/hooks/authHooks/requestOptions.js
@@ -1,12 +1,10 @@
-import { Auth } from "aws-amplify";
+import { fetchAuthSession } from "aws-amplify/auth";
 
 async function requestOptions(body = null) {
   try {
-    const session = await Auth.currentSession();
-    const token = await session.getIdToken().getJwtToken();
-
+    const { idToken } = (await fetchAuthSession()).tokens ?? {};
     const options = {
-      headers: { "x-api-key": token },
+      headers: { "x-api-key": idToken.toString() },
     };
     if (body) {
       options["body"] = body;
--- a/services/ui-src/src/hooks/authHooks/userProvider.js
+++ b/services/ui-src/src/hooks/authHooks/userProvider.js
@@ -1,6 +1,10 @@
 import React, { useCallback, useEffect, useMemo, useState } from "react";
 import { useLocation } from "react-router-dom";
-import { Auth } from "aws-amplify";
+import {
+  signInWithRedirect,
+  signOut,
+  fetchAuthSession,
+} from "aws-amplify/auth";
 import { UserContext } from "./userContext";
 import { AppRoles, IdmRoles } from "../../types";
 import { loadUser } from "../../actions/initial";
@@ -11,7 +15,7 @@ const cartsProdDomain = "https://mdctcarts.cms.gov";
 const tempEndpoint = "https://dt4brcxdimpa0.cloudfront.net";
 
 const authenticateWithIDM = () => {
-  Auth.federatedSignIn({ customProvider: "Okta" });
+  signInWithRedirect({ provider: { custom: "Okta" } });
 };
 
 export const UserProvider = ({ children }) => {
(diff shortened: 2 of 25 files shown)


services/ui-src/src/hooks/authHooks/requestOptions.js
```diff
@@ -1,12 +1,10 @@
-import { Auth } from "aws-amplify";
+import { fetchAuthSession } from "aws-amplify/auth";
 
 async function requestOptions(body = null) {
   try {
-    const session = await Auth.currentSession();
-    const token = await session.getIdToken().getJwtToken();
-
+    const { idToken } = (await fetchAuthSession()).tokens ?? {};
```
benmartin-coforma, line 5: Thought: I'm not a fan of Amplify's typing here. For reference, `fetchAuthSession()` returns
```ts
Promise<{
  tokens?: {
    idToken?: JWT;
    //...
  };
  //...
}>
```
And it feels... catty. "maybe I'll give you tokens, and maaaybe one of them will be an ID token". Which puts us in a position where we have to check for the truthiness of `tokens` and `idToken`. If either is missing, we'll get a runtime error below ("cannot read properties of undefined (reading 'toString')"). To be clear, this is on AWS and not on you.

I think I might prefer an explicit error here over defaulting to undefined, something like
```suggestion
    const apiKey = (await fetchAuthSession()).tokens?.idToken?.toString();
    if (apiKey === undefined) {
      throw new Error("Amplify has betrayed us");
    }
```

But maybe that's going overboard / being too paranoid. If Amplify is being overly... evasive about its types, and the token & idToken are always present, then we will have just added noise to the code, accomplishing nothing

services/ui-src/src/hooks/authHooks/userProvider.js
```diff
@@ -44,8 +48,8 @@ export const UserProvider = ({ children }) => {
 
     // Authenticate
     try {
-      const session = await Auth.currentSession();
-      const payload = session.getIdToken().payload;
+      const { idToken } = (await fetchAuthSession()).tokens ?? {};
+      const { payload } = idToken;
```
benmartin-coforma, line 52: My waffling above applies here too.

services/ui-src/src/util/apiLib.js
```diff
@@ -1,31 +1,36 @@
...
-  return API.get(api, uri, opts);
-};
-const del = (api, uri, opts) => {
-  updateTimeout();
-  return API.del(api, uri, opts);
+const apiRequest = async (request, apiName, path, options) => {
+  try {
+    // swallow request error codes for aws lib, they'll kill the lib
+    await updateTimeout();
+    const response = await request({ apiName, path, options }).response;
```
benmartin-coforma, line 15: I guess it would be too much scope for this PR to hardcode the apiName here, and remove it from all the intermediate functions' parameter lists? I can live with that.

services/ui-src/src/util/apiLib.js
```diff
@@ -1,31 +1,36 @@
...
+    const body = await response?.body?.text(); // body.json() dies on an empty response, spectacularly
+    return body && body.length > 0 ? JSON.parse(body) : null;
```
benmartin-coforma, line 17: > body.json() dies on an empty response

This is very sad, but a good workaround here.

benmartin-coforma (dismissed):
I have not tested, but I approve of the code changes.

services/ui-src/src/util/apiLib.js
BearHanded: I think that's fair, I'll get that done

services/ui-src/src/hooks/authHooks/requestOptions.js
BearHanded: The `?? {}` is their own recommendation for handling it. The whole body.json() thing has a similar vibe. I think I can add a more explicit error here.

benmartin-coforma: Up to you!

benmartin-coforma (approved):
I like it!

gmrabian (approved):
saw no issues in the deployed env. changes look good!

PR #139625, opened 2024-05-15
Reduce possibile e2e failure by swapping "Indicate the range for premiums" to "Indicate the range"

ntsummers1:
### Description

Swapped the text so that you can use either program type for now when filling out the section1 data in regards to the e2e test. Also fixes an id issue with the checkboxes and radio fields. They were all using the same id and creating errors when attempting to swap between checkboxes, or having their id's change to whatever cms design/browser decided it wanted. This caused the DOM to rerender and change the inputs while a user interacted with them. The inputs now all have unique id's for checkboxes. All questions now also attempt to gain their id from either whats being passed to it (Nested Radio fields use this) or whatever the given questions id is.

### Related ticket(s)

CMDCT-3612

---
### How to test

`cd services/ui-src`
`yarn test`

---
### Pre-review checklist


benmartin-coforma (dismissed):
I don't remember writing that comment, but I don't see any reason not to merge in this PR, especially if it helps with e2e failures.

gmrabian (dismissed):

BearHanded (approved):

PR #139644, opened 2024-05-31
Update local seed data with 2023s report

ntsummers1:
### Description

It be doing the title yo

---
### How to test

./run local
See 2023 Alabama report
Enter and see its partially filled and allows you to enter data!

### Pre-review checklist

### DIFF
--- a/services/database/data/seed-local/seed-status.json
+++ b/services/database/data/seed-local/seed-status.json
@@ -1,6 +1,14 @@
 [
   {
     "stateId": "AL",
+    "status": "certified",
+    "year": 2023,
+    "programType": "user",
+    "username": "al@test.com",
+    "lastChanged": "2022-01-04 18:28:18.524133+00"
+  },
+  {
+    "stateId": "AL",
     "status": "in_progress",
     "year": 2022,
     "programType": "combo",
@@ -17,7 +25,7 @@
   },
   {
     "stateId": "AL",
-    "status": "certified",
+    "status": "in_progress",
     "year": 2020,
     "programType": "combo",
     "username": "al@test.com",
--- a/services/database/data/seed-local/seed-status.json
+++ b/services/database/data/seed-local/seed-status.json
@@ -1,7 +1,7 @@
 [
   {
     "stateId": "AL",
-    "status": "certified",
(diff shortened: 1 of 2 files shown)


services/database/data/seed-local/seed-status.json
```diff
@@ -1,4 +1,12 @@
 [
+  {
+    "stateId": "AL",
+    "status": "certified",
```
gmrabian, line 4: the 2023 report shows up as Certified & Submitted when I run locally

gmrabian: Hmm ok. Is that repeatable, so that the test can be re-run and it will still be able to uncertify?
I was mostly addressing this part of the How to test
> Enter and see its partially filled and allows you to enter data!

I'm not able to enter data when it is certified 🤔

ntsummers1: Thats on me. You right. Heres what this PR is intending to do:

Basically the reason the E2E tests were failing in prod vs local is that the types were different. Those types are being set in the seed-local vs seed folders. In local we only had 2022-2021 uncertified and 2020 certified, in the deployed environments we have 2023 certified, 2022-2021 uncertified and 2020 certified. The changes here make dev inline with the seed of the deployed environments. That way if the tests fail we can fix it before it triggers further down the line and doesn't hold up deployments.

What isn't a problem, thankfully, is whether the reports were seeded as certified or not. In CARTS, before an e2e test runs, it calls a function called ensureAvailableReport. That basically goes through and looks for a report thats been certified and uncertifies it to allow for data changes. That way whatever e2e tests that have run before can essentially get reset so theres a clean test bed to work on.

What I'll do to address this:
To make this more clear and dev friendly, I'll swap this (2023 report) to uncertified and leave the old report (2020 report) as certified to handle those test cases. We don't need to test the ensureAvailableReport function, thats tested elsewhere and if it comes it as a problem we can address that in the future.

gmrabian (approved):

services/database/data/seed-local/seed-status.json
gmrabian: Thanks for the context!

PR #139690, opened 2024-07-12
Migrate Questions component

ntsummers1:
### Description

This is the first component being updated as part of [CMDCT-3811](https://jiraent.cms.gov/browse/CMDCT-3811). I'm updating a few things:

- Update the selectIsFormEditable function to use more specific parts of the state
- Migrate mapDispatchToProps to use hooks instead

**There are no tests for this component to test against. This will cause the total test coverage to lower because I added more lines of code. I will be updating this following the dependency upgrades.**

### Related ticket(s)

CMDCT-3811

---
### How to test

Run the application
Play around with all the question types. Yeah, i know, I'm sorry.. This component is used as the base for all questions in the app. Nothing should be different though, the form should update, retain your answers, trigger the header, and let you certify and submit the form.
Admins should be able to log in and see the changes you've made.
Double check that the answers you submit appear the same in Dynamo. It's best to run this as a before/after scenario where you check an input in main and then check in this pr.

---
### Pre-review checklist

---
#### Review


gmrabian (approved):
Didn't see any issues locally

jessabean (approved):

PR #139691, opened 2024-07-15
Convert Text component to hooks

ntsummers1:
### Description

This is the second component being updated as part of [CMDCT-3811](https://jiraent.cms.gov/browse/CMDCT-3811). I'm updating a few things:

- Migrate mapDispatchToProps to use hooks instead

### Related ticket(s)

CMDCT-3811

---
### How to test

Run the application
Play around with the textfields you see in the application. Nothing should be changed. Note this text field is not used to render the text fields underneath radio button choices so no need to test there.
Double check that the answers you submit appear the same in Dynamo. It's best to run this as a before/after scenario where you check an input in main and then check in this pr.

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/components/fields/Text.jsx
+++ b/services/ui-src/src/components/fields/Text.jsx
@@ -1,11 +1,21 @@
 import React, { useEffect, useState } from "react";
-import PropTypes from "prop-types";
+import { useSelector, shallowEqual } from "react-redux";
+//components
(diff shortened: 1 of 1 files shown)


gmrabian (approved):
changes look good. didn't see any data issues locally

jessabean (approved):
👍🏻

PR #139695, opened 2024-07-19
Move SaveError to redux hooks

gmrabian:
### Description

Migrate redux `connect` pattern to `useSelector` hook

If anyone knows how to fix the styling let me know. Right now the error banner runs to the right edge of the window.

### Related ticket(s)

CMDCT-3814

---
### How to test

https://def61orvrftr6.cloudfront.net/
- Log in as a state user
- Enter any in progress report
- Change something and verify it saves
- Turn your wifi off
- Change something on the page and verify it throws an error when it tries to save
- Turn your wifi back on and verify the error disappears on its own once a save succeeds

---
### Pre-review checklist

### DIFF
--- a/services/ui-src/src/components/fields/Autosave.jsx
+++ b/services/ui-src/src/components/fields/Autosave.jsx
@@ -1,4 +1,3 @@
-import React from "react";
 import { Button } from "@cmsgov/design-system";
 
 const Autosave = () => (
--- a/services/ui-src/src/components/fields/Autosave.jsx
+++ b/services/ui-src/src/components/fields/Autosave.jsx
@@ -1,3 +1,4 @@
+import React from "react";
 import { Button } from "@cmsgov/design-system";
 
 const Autosave = () => (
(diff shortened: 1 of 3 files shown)


britt-mo (commented):
Tested, functionality was as expected. I think the SaveError margins need tweaking in `_alerts.scss` 

```
  &.alert--unexpected-error__active {
    top: 60px;
    margin-top: 0;
  }
```

bangbay-bluetiger (dismissed):
LGTM. I think the banner would look better as 100% width.

gmrabian:
@bangbay-bluetiger I tried 100% width and it ran further off the page 😅. Maybe I targeted the wrong element.

@ntsummers1 I gave that a shot and the styling is better. This is the only component that uses these css classes from my search. It had some visual conflicts with the account dropdown but so does the original. If you are ok with this we can go ahead with this new style.
Screenshot 2024-07-19 at 11 27 59 AM

ntsummers1:
@gmrabian Works for me! The dropdown could go over top if we gave it a z-index of 10 or something, but I think its fine. Unless product says otherwise I think this is an overall improvement :)

ntsummers1 (approved):

britt-mo (approved):

PR #139696, opened 2024-07-19
Update Userinfo component

gmrabian:
### Description

Migrate redux `connect` pattern to `useSelector` hook

### Related ticket(s)

CMDCT-3812

---
### How to test

Run `main` locally
- Log in as a state user
- Navigate to `/userinfo`
- Check the page at different widths
- Check the console for errors
- Change to this branch `cmdct-3812` and verify there are no changes to the page

Alternatively, compare the [deployed branch](https://d2k3tvv8jlodnc.cloudfront.net/) with [CARTS dev](https://mdctcartsdev.cms.gov/)

### Notes

I tried to rename this file to `UserInfo` but the AppRoutes tests started failing on the import 🤔 

---
### Pre-review checklist


services/ui-src/src/components/sections/Userinfo.jsx
```diff
@@ -18,20 +21,14 @@ const Userinfo = ({ currentUser }) => {
     );
   });
   return (
-    <ul>
-      {info.map((item) => (
-        <li>{item}</li>
-      ))}
-    </ul>
+    <main className="main">
```
bangbay-bluetiger, line 24: Why the addition of `main` element?

gmrabian: The accessibility test threw an error that the page didn't have any landmarks. Granted, the test component doesn't have the header and footer, etc.
I can remove it. Not sure the best way to handle the accessibility test

services/ui-src/src/components/sections/Userinfo.jsx
```diff
@@ -1,8 +1,11 @@
 import React from "react";
-import PropTypes from "prop-types";
-import { connect } from "react-redux";
+import { shallowEqual, useSelector } from "react-redux";
 
-const Userinfo = ({ currentUser }) => {
+const Userinfo = () => {
```
bangbay-bluetiger, line 4: You can rename the file to `UserInfo` using `git mv Userinfo.jsx UserInfo.jsx` and after you edit `AppRoutes`, the error seems to go away after a TypeScript restart.

services/ui-src/src/components/sections/Userinfo.jsx
```diff
@@ -18,20 +21,14 @@ const Userinfo = ({ currentUser }) => {
...
   return (
-    <ul>
-      {info.map((item) => (
-        <li>{item}</li>
-      ))}
-    </ul>
+    <main className="main">
+      <ul>
+        {info.map((item) => (
+          <li>{item}</li>
```
bangbay-bluetiger, line 27: `<li>` needs a `key` prop.

services/ui-src/src/components/sections/Userinfo.jsx
bangbay-bluetiger: It seems `main` should be defined in a higher level component rather than here, but it also seems more trouble than necessary at the moment.

services/ui-src/src/components/sections/Userinfo.jsx
```diff
@@ -1,8 +1,11 @@
...
-import PropTypes from "prop-types";
-import { connect } from "react-redux";
+import { shallowEqual, useSelector } from "react-redux";
 
-const Userinfo = ({ currentUser }) => {
+const Userinfo = () => {
+  const currentUser = useSelector(
+    (state) => state.stateUser.currentUser,
+    shallowEqual
+  );
```
ntsummers1, line 8:
```suggestion
  const currentUser = useSelector((state) => state.stateUser.currentUser);
```

This can just be a straight call, I've been using the shallowEqual only so I don't have to write out multiple useSelector calls and can combine them into one thing.

ntsummers1 (commented):
Preciate you for writing a test when you didn't even have to <3

services/ui-src/src/components/sections/Userinfo.jsx
gmrabian: Good to know! I should have looked at what it was doing instead of just copy-pasting 😅

services/ui-src/src/components/sections/Userinfo.jsx
gmrabian: Thanks! This did the trick

services/ui-src/src/components/sections/Userinfo.jsx
gmrabian: You're right. I think I'll remove the main element and the accessibility test for now. Fudging the test doesn't seem like a good solution and since there weren't any tests to begin with it's still an improvement to have one.

services/ui-src/src/components/sections/UserInfo.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+const UserInfoComponent = (
+  <Provider store={stateUser}>
+    <UserInfo />
+  </Provider>
+);
+
+describe("UserInfo", () => {
+  it("should render the UserInfo Component correctly", () => {
+    expect(shallow(UserInfoComponent).exists()).toBe(true);
```
bangbay-bluetiger, line 22: I think we can add tests that give have more meaningful coverage. How about:
```
const wrapper = mount(UserInfoComponent);
expect(wrapper.find("li").first().text()).toBe("username: stateuser2@test.com");
expect(wrapper.find("li").at(1).text().trim()).toBe("state: AL");
```

gmrabian: Thank you! I was having some trouble with `getByText` and `queryAllByText` for some reason but targeting the list items here did the trick 🪄

ntsummers1 (approved):

jessabean (approved):

PR #139702, opened 2024-07-22
Migrate SynthesizedTable+SynthesizedValue

ntsummers1:
### Description

These is the seventh and eighth component being updated as part of [CMDCT-3811](https://jiraent.cms.gov/browse/CMDCT-3811). I'm updating a few things:

- Migrate mapStateToProps to use hooks instead
- Fix up all synthesize value calls to use specific parts of the state instead of a global check 

Because both of these components are so heavily tied to the synthesize call which I've touched in inputs for, I've updated them both here.

**Note: This lowers test coverage because there were no original tests for these components. Will be adding later with others after dependency updates**

Synthesized Table:

https://github.com/user-attachments/assets/f4e28795-53b1-4488-8bd6-a96a1f7cddeb

Synthesized Value:

https://github.com/user-attachments/assets/279dc5c6-1b43-439a-bd43-c89ecfe7cfa0

### Related ticket(s)

CMDCT-3811

---
### How to test

Run the application
Play around with the Synthesize Table and Synthesize value component. They should auto update as you modify the data that they are associated with!

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/components/fields/SynthesizedTable.jsx
+++ b/services/ui-src/src/components/fields/SynthesizedTable.jsx
@@ -107,7 +107,7 @@ const SynthesizedTable = ({ question, tableTitle }) => {
 };
 SynthesizedTable.propTypes = {
   question: PropTypes.object.isRequired,
-  title: PropTypes.string.isOptional,
+  tableTitle: PropTypes.string.isOptional,
 };
 
 export default SynthesizedTable;
--- a/services/ui-src/src/components/fields/SynthesizedTable.jsx
+++ b/services/ui-src/src/components/fields/SynthesizedTable.jsx
@@ -18,37 +18,21 @@ const SynthesizedTable = ({ question, tableTitle }) => {
       shallowEqual
     );
 
-  const calculateRows = (
-    allStatesData,
-    stateName,
-    stateUserAbbr,
-    chipEnrollments,
-    formData
-  ) =>
-    question.fieldset_info.rows.map((row) =>
-      row.map((cell) => {
-        const value = synthesizeValue(
-          cell,
-          allStatesData,
-          stateName,
-          stateUserAbbr,
-          chipEnrollments,
-          formData
-        );
+  const rows = question.fieldset_info.rows.map((row) =>
+    row.map((cell) => {
+      const value = synthesizeValue(
+        cell,
+        allStatesData,
+        stateName,
+        stateUserAbbr,
+        chipEnrollments,
+        formData
+      );
 
-        return typeof value.contents === "number" &&
-          Number.isNaN(value.contents)
-          ? { contents: "Not Available" }
-          : value;
-      })
-    );
-
-  const rows = calculateRows(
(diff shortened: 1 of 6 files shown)


services/ui-src/src/components/fields/SynthesizedTable.jsx
```diff
@@ -1,9 +1,56 @@
...
+      })
+    );
+
+  const rows = calculateRows(
+    allStatesData,
+    stateName,
+    stateUserAbbr,
+    chipEnrollments,
+    formData
+  );
```
gmrabian, line 52: nit: does this need to be separate from the function `calculateRows` if we only invoke it once? What if we just assigned rows as such
```
const rows = question.fieldset_info.rows.map...etc.
```

gmrabian (dismissed):

services/ui-src/src/components/fields/SynthesizedTable.jsx
```diff
@@ -1,9 +1,56 @@
...
+        const value = synthesizeValue(
+          cell,
+          allStatesData,
+          stateName,
+          stateUserAbbr,
+          chipEnrollments,
+          formData
+        );
+
+        return typeof value.contents === "number" &&
```
bangbay-bluetiger, line 39: How can a value that is typeof `number` also be `NaN`?

ntsummers1: From what I can tell, the data is stored not as a number but as a string, and the that value can be '7' or something like ''. Definitely something I'd like to fix, though a current proposal is having some numbers come through as '<11' so until the details on that are ironed out I hesitate to mess with anything here.

services/ui-src/src/components/fields/SynthesizedTable.jsx
ntsummers1: Done! :D

services/ui-src/src/components/fields/SynthesizedTable.jsx
bangbay-bluetiger: It seems like a scenario that could never be true, because if `value.contents = "7"`, `typeof` would return `string` and not `number`. Instead, we could cast the value then check the number like this:
```
Number.isNaN(parseInt(value.contents, 10))
```

ntsummers1: Screenshot 2024-07-25 at 3 21 35 PM

Well I found the true scenario. When reading in data from past forms, it does indeed read the values as a number. However, sometimes those values that are pulled down are literally NaN which triggers this use case. So it comes out into a table looking like this:

Screenshot 2024-07-25 at 3 22 38 PM

I was not able to replicate this with my own user inputted data, it had to come from a prior form. 

I think this is a little messed up, though I need to work with the PM and BO's before I touch this legacy logic. Maybe this was truly intended, maybe not. But I will open an investigation outside of this PR to look into this.

gmrabian (approved):

PR #139727, opened 2024-08-01
remove references to branch master

gmrabian:
### Description

The `master` branch of CARTS holds what we call CARTS v2. It uses a different pattern and infrastructure than the rest of our MDCT suite. Two years ago we migrated to `main` (CARTS v3). All of our dev infrastructure and code promoted to prod comes from `main`

`master` has been copied to `skipci-archive-master`

We need to remove any references to master in the code so as to avoid errors and confusion.

Once this is merged, github settings are updated, and we've determined it's safe to do so, we will delete the branch named `master`

### Related ticket(s)

CMDCT-3804

---
### How to test

See if you can find any more references that need to be updated.

Verify CARTS works as expected https://d13roq1oz0l6c0.cloudfront.net/

### Notes

Someone with higher permissions will need to remove branch protections and any other git configuration on master

🤔 can you think of a better archived branch name? `skipci-archive-carts-v2`? Also how long do we anticipate keeping this, and for what purpose? I imagine the drift is already large enough that it may not be helpful to reference the old master branch.

---
### Pre-review checklist

- [ ] ~I have added [thorough](https://shorturl.at/aejkF) tests, if necessary~

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![Maintainability](https://api.codeclimate.com/v1/badges/f1775f53aedf747e85b2/maintainability)](https://codeclimate.com/repos/6449718c21275100df510ea9/maintainability)
 [![Test Coverage](https://api.codeclimate.com/v1/badges/f1775f53aedf747e85b2/test_coverage)](https://codeclimate.com/repos/6449718c21275100df510ea9/test_coverage)
 
-### Integration Environment Deploy Status:
+## Integration Environment Deploy Status:
 | Branch  | Build Status |
 | ------------- | ------------- |
 | main  | ![deploy](https://github.com/Enterprise-CMCS/macpro-mdct-carts/actions/workflows/deploy.yml/badge.svg)  |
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -115,10 +115,7 @@ On the SEDS side, this topic is updated on every submission of seds data, but CA
 - 4th quarter data.
 - The rollover for a "new year" is October, and future submissions are not recognized until that threshold
 
-Updates outside of that time frame will need to be manually corrected in CARTS, or the integration will need to be modifed to collect data for old forms. CARTS additionally looks for the `enrollmentCounts` property which is only included in forms 21E and 64.21E (question 7), either by manual trigger or update. See SEDS files:
-
-- [generateEnrollmentTotals](https://github.com/Enterprise-CMCS/macpro-mdct-seds/blob/master/services/app-api/handlers/state-forms/post/generateEnrollmentTotals.js)
-- [updateStateForms](https://github.com/Enterprise-CMCS/macpro-mdct-seds/blob/master/services/app-api/handlers/state-forms/post/updateStateForms.js)
+Updates outside of that time frame will need to be manually corrected in CARTS, or the integration will need to be modifed to collect data for old forms. CARTS additionally looks for the `enrollmentCounts` property which is only included in forms 21E and 64.21E (question 7), either by manual trigger or update.
```
gmrabian, line 118: are there new files in `main` that we want to reference here?

ntsummers1: I don't think so.. @BearHanded? I don't think theres anything new related to this, right?

dwhitestratiform (approved):

gmrabian:
> I think the `master` in `.github/workflows/destroy.yml` is referencing the master branch of `Enterprise-CMCS/bigmac`, and not of this repo - is that correct?

Correct. The [documentation](https://github.com/benc-uk/workflow-dispatch?tab=readme-ov-file#ref) states we need to specify ref to ensure we use the master branch of that action, rather than have GitHub infer our branch name for the same.

gmrabian:
> I don't have any problem with a branch named `skipci-archive-master`. I do think `skipci-archive-carts-v2` may be a better choice, because it's more direct; it doesn't require you to remember the factoid "we switched from master to main when we rewrote the app and called it v3". Although it does require you to remember the factoid "the current incarnation of CARTS is called v3" 🤔
> 
> As for how long we would keep it around, I think the answer is "forever", right? I mean, it's a large part of the history of the project, and that's a large part of what Git is for.

I asked for feedback on the branch name for the same reasons you mentioned. I think it might be good to keep `master` out of it if possible. Maybe `skipci-archive-carts-old`? Not a big deal regardless.

Good point about Git history

PR #139730, opened 2024-08-06
Convert Title to use hooks

ntsummers1:
### Description

This migrates the Title component. Specifically, this changes the following:

Removes the connect property
Uses hooks to fetch state

Screenshot 2024-08-12 at 12 23 49 PM

### Related ticket(s)

CMDCT-3813

---
### How to test

Load up a report
See the title on the top of the page! Its always `{StateName} CARTS FY{formYear} Report`

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/components/layout/Title.jsx
+++ b/services/ui-src/src/components/layout/Title.jsx
@@ -1,8 +1,16 @@
 import React from "react";
+import { useSelector, shallowEqual } from "react-redux";
+//types
 import PropTypes from "prop-types";
-import { connect } from "react-redux";
 
-const Title = ({ name, stateName, formYear, urlStateName }) => {
+const Title = ({ urlStateName }) => {
+  const [name, stateName, formYear] = useSelector(
+    (state) => [
+      state.stateUser.name.state.global.stateName,
+      state.global.formYear,
+    ],
+    shallowEqual
+  );
   const displayStateName = name || urlStateName || stateName || "";
 
   return (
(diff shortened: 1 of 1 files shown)


services/ui-src/src/components/layout/Title.jsx
```diff
@@ -14,16 +23,7 @@ const Title = ({ name, stateName, formYear, urlStateName }) => {
   );
 };
 Title.propTypes = {
-  name: PropTypes.string,
-  stateName: PropTypes.string,
-  urlStateName: PropTypes.string,
```
gmrabian, line 19: I think you accidentally deleted `urlStateName` instead of `formYear`

gmrabian (changes requested):
One issue with proptypes but still works — do you know if the proptypes still apply when not using connect?

services/ui-src/src/components/layout/Title.jsx
ntsummers1: You are 100% correct. Good catch!

ntsummers1:
> One issue with proptypes but still works — do you know if the proptypes still apply when not using connect?

They do, though it seems they mainly serve as assistance for when you hover over a component. For instance, hovering over the `Title` in the code when referencing the Title Component will show that it takes a prop type of urlStateName which is a string. It doesn't seem to enforce the types though, which begs the question if its truly that helpful or do we remove PropTypes all together and just let native Typescript take control here.

gmrabian (approved):

britt-mo (approved):
Tested locally, title appeared as expected

PR #139749, opened 2024-08-20
prevent state name from extending past the sidebar column

britt-mo:
### Description
The state name was extending past the sidebar column and at times overlapping with the report text:

Screenshot 2024-07-26 at 10 16 39 AM

This change will make the state name drop below the state icon if the name exceeds the column width.

### Related ticket(s)

CMDCT-3876

---
### How to test
Login, enter a report, play around with the state name. use devtools to change the state name to different lengths and resize the page to see the ways the text moves.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/layout/StateHeader.jsx
+++ b/services/ui-src/src/components/layout/StateHeader.jsx
@@ -1,20 +1,25 @@
-import React from "react";
 import { useSelector } from "react-redux";
+import { AppRoles } from "types";
 
 const StateHeader = () => {
-  const { name, imageURI } = useSelector((state) => state.stateUser);
-
+  const { currentUser, name, imageURI } = useSelector(
+    (state) => state.stateUser
+  );
   return (
-    <div
-      className="state-header"
-      data-testid="state-header"
-      aria-label="State Header"
-    >
-      <div className="state-image">
-        <img src={imageURI} alt={name} />
-      </div>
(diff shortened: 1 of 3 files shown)


gmrabian (changes requested):
Blocking due to style change not expected in ticket

The admin view has a gap now (also did admin always not have the state name and logo?)

https://github.com/user-attachments/assets/b7ae604b-9632-4b1f-b3c7-597a5f692dc6

ntsummers1 (commented):
Looking much better! My only question is what should we do with states with longer continuous names, specifically something like Massachusetts. Right now it'll do the same thing:

Screenshot 2024-08-22 at 9 54 37 AM

We could trim it to the max width of the sidebar, so something like this: 

Screenshot 2024-08-22 at 9 51 42 AM

or even add an elipsis:

Screenshot 2024-08-22 at 10 21 15 AM

I'm not in love with either of these options though, so I'm curious on your thoughts!

britt-mo:
@ntsummers1 I would love to shrink the state name text size itself to just fit the container...not sure if that's an option!

gmrabian (approved):
Text wrap looks good! admin spacing is correct too

Screenshot 2024-08-22 at 4 06 01 PM

ntsummers1 (approved):

PR #139755, opened 2024-08-28
remove momentjs

gmrabian:
### Description

Replace momentjs with `date-fns` and native Date type

the moment package is a sub-dependency of `serverless-bundle`. we cannot remove it from yarn.lock, but hopefully this still resolves the issue.

References:
[Security issue](https://www.invicti.com/web-vulnerability-scanner/vulnerabilities/version-disclosure-momentjs/)
[MomentJS recommends using alternatives](https://momentjs.com/docs/#/-project-status/)
[date-fns docs](https://date-fns.org/v3.6.0/docs/Getting-Started)

### Related ticket(s)

CMDCT-3906

---
### How to test

Tests pass

Timeout
- Modify the values for `IDLE_WINDOW` and `PROMPT_AT` in the `services/ui-src/src/hooks/authHooks/authLifecycle.js` file to be shorter (change leading values to 1 and 0.5 or something)
- Run locally
- Let the app sit until the timeout hits
  - Ensure the prompt shows up as expected
  - Ensure the prompt still does the correct actions when you click on them
  - Ensure if you don't do anything it logs you out

Save message
- Enter any report as a state user and change something
- Verify the autosave banner switches from "Not yet saved" to "Saved"
- Verify that if you let it sit for 1+ minutes the message changes to something like "Last saved 9:30 AM (5 minutes ago)"

Thanks message
- Enter any report as a state user and submit it
- Verify the message shows the date and time formatted correctly
  - "Submitted on August 28th, 2024 at 9:35 AM by Frank States."
- Verify that if you refresh the page it still shows the correct info

Page info
- Enter any report as a state user
- Verify that if you make a change the section at the top that says "Draft | Last Edited ... " shows the correct date

### Notes

Our SaveMessage had cases for showing a message if the last save was multiple days or years ago, but our auth times out after 30 minutes. I have doubts that our state users would stay up overnight to stay logged in, so I think it's safe to remove those cases.

---
### Pre-review checklist


services/ui-src/src/components/layout/CertifyAndSubmit.jsx
```diff
@@ -43,15 +43,25 @@ const Submit = ({ openCertifyConfirmation }) => (
...
+  const formattedTime = lastSave ? format(lastSave, "p") : "";
   return (
     <>
       <h3 data-testid={"certifyThankYou"}>
         Thank you for submitting your CARTS report!
       </h3>
       <p>
-        Submitted on {lastSave.format("MMMM Do, YYYY")} at{" "}
-        {lastSave.format("h:mm A")} by {submitterUsername}.
+        Submitted on {formattedDate} at {formattedTime} by {submitterUsername}.
```
bangbay-bluetiger, line 64: If `lastSave` has no value, it seems strange to print this out because it'll look like `Submitted on at by user`. What if instead we have something like `{submittedText}` and then:

```
let submittedText = `Submitted by ${submitterUsername}.`;

if (lastSave) {
  const formattedDate = format(lastSave, "PPP");
  const formattedTime = format(lastSave, "p");

  submittedText = `Submitted on ${formattedDate} at ${formattedTime} by ${submitterUsername}.`
}
```

Or maybe show nothing if there's no `lastSave`?

bangbay-bluetiger (approved):
Suggestion, but not a blocker.

ntsummers1 (approved):
Like the small move of the the lastSave selector to the Thanks component in CertifyAndSubmit. Nice work!

services/ui-src/src/components/layout/CertifyAndSubmit.jsx
gmrabian: If we don't have `lastSave` it's likely we won't have `submitterUsername` as well since they source from the same function, `getCurrentReportStatus`, so no text is probably the best alternative.

That said the functionality I'm seeing right now is that when you refresh the page it shows the unsubmitted text until it retrieves the status, at which point it shows the Thanks text. The status is sourced from the same place as `lastSave` and `submitterUsername`: `getCurrentReportStatus` ([relevant selectors](https://github.com/Enterprise-CMCS/macpro-mdct-carts/blob/main/services/ui-src/src/store/selectors.js#L183-L236))

This "Submitted on..." text can only show up if `getCurrentReportStatus` returns a non-empty object. Since that object contains all of the other information we are looking for, our main concern then is if the database houses all the expected information, meaning did we save all of it.

I think the change is fine, though I think our other code logic makes it highly unlikely

PR #139767, opened 2024-09-18
BUGFIG: Remove parseInt() for previous year value to allow non-numbers to display as empty

gmrabian:
### Description

This bug showed up where null or empty string values were rendered as `NaN` in their cells

These values should (in priority order)
1. render user entered values
2. pull from previous year
3. show up blank if neither of the above

### Related ticket(s)

CMDCT-3915 cont.

---
### How to test

[CARTS dev](https://mdctcartsdev.cms.gov/)
- Log in as stateuser2@test.com
- Enter AL 2021
- Go to section 3c, part 5
- Select “No” to the question 2
- Verify many values display as NaN

https://d2z1dvefrowpqq.cloudfront.net/
- Repeat the above steps
- Verify the previously NaN cells show up as empty cells
- Go to AL 2022, Section 3c, part 5
- Enter some numbers, delete some numbers
- Verify that in AL 2023 those show up as numbers or blank as appropriate
- Verify user entered values for each year take precedence over last year

### Notes

This is how the Datagrid logic was avoiding the NaN problem—it never looked for numbers to begin with!

---
### Pre-review checklist

---
#### Review
---

### DIFF
--- a/services/ui-src/src/components/fields/Integer.jsx
+++ b/services/ui-src/src/components/fields/Integer.jsx
@@ -41,7 +41,7 @@ const getPrevYearValue = (question, lastYearFormData) => {
 
     // The first will always be correct
     if (matchingQuestion[0]) {
-      prevYearValue = parseInt(matchingQuestion[0].questions[0].answer?.entry);
+      prevYearValue = matchingQuestion[0].questions[0].answer?.entry;
     }
   }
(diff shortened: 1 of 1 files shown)


ntsummers1 (approved):
It be fixing the NaN!

PR #139775, opened 2024-09-24
[CMDCT-4013] Hide Some Calculated Values in Print View

ntsummers1:
### Description

This PR shows the ability to hide certain calculated table columns and calculated values from the user in the print view of the application. Specifically, this tackles only the 2024 report’s Section 3C Part 2 and Part 3 tables and Section 3C Part 3 calculated values. If this looks like its working as intended, I’ll extend this across all of the parts and calculated values and update the 2024 base template accordingly

For a better look at what has changed, heres the normal view of Section 3C Part 2 and Part 3

Normal View:

Screenshot 2024-09-24 at 12 38 44 PM

Screenshot 2024-09-24 at 12 39 14 PM

Print View:

Screenshot 2024-09-24 at 12 38 56 PM

Screenshot 2024-09-24 at 12 39 29 PM

### Related ticket(s)

CMDCT-4013

---
### How to test

Open the local 2024 report as stateuser2.
Navigate to Section 3c Part 2
Note that the table shows Number and percentage
Enter some data if you want and see it calculates
Scroll down to the print button and click print section
Navigate to Part 2 and see that the number column has been hidden
Back on the normal view of the page, navigate to Navigate to Section 3c Part 3
See that theres a computed value after question 4 as well as 2 more tables
Enter data and see it works
Scroll down to the print button and click print section
Navigate to Part 3 and see that the number column has been hidden in both tables and the calculated value is also hidden.

---
### Pre-review checklist

---
#### Review


benmartin-coforma (approved):
I haven't tested this, but it looks like a clean solution! Now that we're at four instances of `"lessThanEleven"` in the code, it may be time to consider declaring it in your constants.js file? Up to you though 👉😎👉

britt-mo (commented):
Tested and everything worked as described, but I also have a question -- do we care if users input data that will result in percentages over 100%?

ntsummers1:
> Tested and everything worked as described, but I also have a question -- do we care if users input data that will result in percentages over 100%?

We do not. Is a good call out though as I don't think it should ever be the case that these values should reach over 100%, but as of this time we don't have any validation or direction from the BO's to ensure these calculations are at or below 100%

gmrabian (approved):
Britt raises a good question, though I know that our existing functionality allows for percentages greater than 100%

ntsummers1:
> I haven't tested this, but it looks like a clean solution! Now that we're at four instances of `"lessThanEleven"` in the code, it may be time to consider declaring it in your constants.js file? Up to you though 👉😎👉

Made an enum to contain it and all future masks (Should we require them :D)

karla-vm (approved):

PR #139787, opened 2024-10-02
Remove section 1's question 9 accidental carryover

ntsummers1:
### Description

We found that Section 1 Question 9 was not supposed to be in the 2023 report. We've thus removed it from the upcoming 2024 report. I also went through and did some quick typo fixes, as well as properly added N/A to the Parts 3 and 4 Questions that were being added at run time.

---
### How to test

In the deployed instance, checkout the 2024 report
See that the old question 9 was removed and the questions following have been moved down and taken its place.
See that N/A is in all the radio questions of Parts 3 and 4 in Section 1
Checkout the 2023 Report
See that the old question 9 is still there
See that N/A is in all the radio questions of Parts 3 and 4 in Section 1

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/database/data/seed/seed-section-base-2024.json
+++ b/services/database/data/seed/seed-section-base-2024.json
@@ -704,6 +704,10 @@
                         {
                           "label": "No",
                           "value": "no"
+                        },
+                        {
+                          "label": "N/A",
+                          "value": "n/a"
                         }
                       ]
                     },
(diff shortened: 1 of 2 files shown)


gmrabian (commented):
one small issue that the number shows up with a dash in front for 2024, but otherwise looks good
Screenshot 2024-10-02 at 10 50 35 AM

ntsummers1:
> one small issue that the number shows up with a dash in front for 2024, but otherwise looks good 

Good catch! Fixed :D

gmrabian (approved):

PR #139800, opened 2024-10-18
Print section fix for additional user types

ntsummers1:
### Description

In [CMDCT-3984](https://jiraent.cms.gov/browse/CMDCT-3984) we fixed it so that admins could see the print section view. I didn't realize, however, that there are additional user types that have to get the specific URL type. Therefor I've updated the list so that those users dont have the site crash when attempting to print a particular form section.

---
### How to test

Using cms.admin@test.com, internaluser@test.com, help.desk@test.com, stateuser2@test.com, help.approver@test.com, and cms.user@test.com, login to carts and enter a report. 
Click on the print button on the bottom and click this section
See that the page loads correctly
Back out and click on the print button on the bottom and click this form
See that the page loads correctly

---
### Pre-review checklist

### DIFF
--- a/services/ui-src/src/components/layout/FormActions.jsx
+++ b/services/ui-src/src/components/layout/FormActions.jsx
@@ -27,13 +27,7 @@ const FormActions = () => {
   let sectionId = "";
 
   const role = currentUser.role;
-  if (
-    role !== AppRoles.CMS_ADMIN &&
-    role !== AppRoles.INTERNAL_USER &&
-    role !== AppRoles.HELP_DESK &&
-    role !== AppRoles.CMS_APPROVER &&
-    role !== AppRoles.CMS_USER
-  ) {
+  if (role == AppRoles.STATE_USER ) {
     searchParams = document.location.pathname
       .toString()
       .replace("/sections/", "")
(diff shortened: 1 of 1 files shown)


britt-mo (dismissed):
I should have thought about this!! Working as expected now

services/ui-src/src/components/layout/FormActions.jsx
```diff
@@ -26,21 +26,28 @@ const FormActions = () => {
...
 
-  if (currentUser.role === AppRoles.CMS_ADMIN) {
-    const stateId = window.location.href.split("/")[5];
+  const role = currentUser.role;
+  if (
+    role !== AppRoles.CMS_ADMIN &&
+    role !== AppRoles.INTERNAL_USER &&
+    role !== AppRoles.HELP_DESK &&
+    role !== AppRoles.CMS_APPROVER &&
+    role !== AppRoles.CMS_USER
```
gmrabian, line 35: can we just set it for `if role === AppRoles.STATE_USER` ? Are we likely to add other roles that will have the same url format as a state user?

gmrabian (dismissed):
looks good; works as expected.

I think we can be less verbose on the if statement

services/ui-src/src/components/layout/FormActions.jsx
ntsummers1: 💯

PR #139844, opened 2024-11-25
Section 5 Part 2 Table Fix

ntsummers1:
### Description

It looks like theres a possibility that sometimes the values users enters will get commas added to them. When these commas get added, they will hard break a sum calculation. You can see that here:

https://github.com/user-attachments/assets/06ca7a9c-fe29-476f-bdc3-7b66c018f3f0

This is a quick fix, simply removing the comma's when the value gets calculated. You can see that in the video below. However, this is **_a realllllllllly big problem_**. Data integrity is all messed up here and based on the comments in other functions, looks like this is known too. Will create a ticket to manage and actually fix this blatant issue.

After Fixing

https://github.com/user-attachments/assets/ce6ce420-039e-4197-bc19-cd836995bd91

### Related ticket(s)

Support Ticket #CS2388324

---
### How to test

Login as stateuser2 (Important)
Open the 2024 report
Navigate to Section 5, Enter some data into Part 2
On a new terminal, run `DYNAMO_ENDPOINT=http://localhost:8000 dynamodb-admin`
Open he seed-section, and find Alabama 2024 Section 5
Look for the field `2024-05-a-02-01-a`.
Add a comma to its value
Save the dynamo change (Top left of the page)
Hard refresh the local instance of CARTS
See that it don't break in the table and calculated despite your comma addition!

(If you're confused, see the first video on how I did this)

---
### Pre-review checklist

---
### Review

### DIFF
--- a/services/ui-src/src/util/synthesize.js
+++ b/services/ui-src/src/util/synthesize.js
@@ -177,8 +177,10 @@ const formula = (targets, providedFormula, precision) => {
 const sum = (values) => {
   let returnValue = "";
   const hasNumbers = values.some((value) => value !== null && value !== "");
+
   if (hasNumbers) {
-    returnValue = values.reduce((acc, value) => acc + +value, 0);
+    const cleanedValues = values.map((value) => value.replace(/,/g, ""));
+    returnValue = cleanedValues.reduce((acc, value) => acc + +value, 0);
   }
   return returnValue;
 };
--- a/services/ui-src/src/util/synthesize.js
+++ b/services/ui-src/src/util/synthesize.js
@@ -179,7 +179,9 @@ const sum = (values) => {
   const hasNumbers = values.some((value) => value !== null && value !== "");
 
   if (hasNumbers) {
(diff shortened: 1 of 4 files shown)


services/ui-src/src/util/synthesize.js
```diff
@@ -177,8 +177,12 @@ const formula = (targets, providedFormula, precision) => {
 const sum = (values) => {
   let returnValue = "";
   const hasNumbers = values.some((value) => value !== null && value !== "");
+
   if (hasNumbers) {
-    returnValue = values.reduce((acc, value) => acc + +value, 0);
+    const cleanedValues = values.map(
```
BearHanded, line 182: can you do something where you cast to a number rather than relying on a regex?

jessabean (approved):

services/ui-src/src/util/synthesize.js
ntsummers1: Unfortunately not :(

PR #139867, opened 2024-12-17
Upgrade kafkajs package

gmrabian:
### Description

- Upgrade kafkajs package to major version 2
- Rename unmarshall function for clarity
  - although we could just use this function directly, I think the wrapper at least allows future updates to be easier. I believe when we went from v2 to v3 in aws-sdk we had to remove a configuration parameter on the unmarshaller

Resources:
[Release log](https://github.com/tulios/kafkajs/releases)
[Official migration guide](https://kafka.js.org/docs/migration-guide-v2.0.0)

### Related ticket(s)

CMDCT-4174

---
### How to test

- Perform actions that update the DB
- Check the kafka lambda logs to ensure they update (`carts-bigmac-streams-cmdct-4174-postKafkaData`)

### Notes

I matched functionality from v1. For example: timeouts are now default, but I disabled them in our config to match what we did before `enforceRequestTimeout: false`

The new functionality allows headers to be arrays, and expects us to check for them. Nowhere in our code do we use `.headers` so I think we are ok. I verified our lambda logs can process arrays of messages as well.

---
### Pre-review checklist

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -161,10 +161,9 @@ async function deploy(options: { stage: string }) {
   const deployCmd = ["sls", "deploy", "--stage", stage];
   await runner.run_command_and_output("SLS Deploy", deployCmd, ".");
   // Only deploy resources for kafka ingestion in real envs
-
-  // if (stage === "main" || stage === "val" || stage === "production") {
-  await deploy_kafka_service(runner, stage);
-  // }
+  if (stage === "main" || stage === "val" || stage === "production") {
+    await deploy_kafka_service(runner, stage);
+  }
 }
 
 async function destroy_stage(options: {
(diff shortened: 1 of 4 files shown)


src/run.ts
```diff
@@ -161,9 +161,10 @@ async function deploy(options: { stage: string }) {
   const deployCmd = ["sls", "deploy", "--stage", stage];
   await runner.run_command_and_output("SLS Deploy", deployCmd, ".");
   // Only deploy resources for kafka ingestion in real envs
-  if (stage === "main" || stage === "val" || stage === "production") {
-    await deploy_kafka_service(runner, stage);
-  }
+
+  // if (stage === "main" || stage === "val" || stage === "production") {
+  await deploy_kafka_service(runner, stage);
+  // }
```
gmrabian, line 167: TODO: uncomment protections after reviews

bangbay-bluetiger (dismissed):
Verified log changes.

ntsummers1 (dismissed):

bangbay-bluetiger (approved):

PR #139874, opened 2025-01-02
Create create-contact-list database script

JonHolman:
### Description
New database script to create a contact list CSV file

### Related ticket(s)

CMDCT-

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/database/scripts/create-contact-list.js
+++ b/services/database/scripts/create-contact-list.js
@@ -62,7 +62,7 @@ function transform(items) {
   return items.map((item) => {
     const subsections = item.contents.section.subsections?.[0];
     const parts = subsections.parts?.[0];
-    const questions = parts.questions;
+    const { questions } = parts;
 
     const stateName = questions[0].answer.entry;
     const contactDetails = questions[3].questions;
--- a/services/database/scripts/create-contact-list.js
+++ b/services/database/scripts/create-contact-list.js
@@ -1,0 +1,96 @@
+/* eslint-disable no-console */
+/*
+ * Local:
+ *    `DYNAMODB_URL="http://localhost:8000" dynamoPrefix="local" node services/database/scripts/create-contact-list.js`
+ *  Branch:
+ *    dynamoPrefix="YOUR BRANCH NAME" node services/database/scripts/create-contact-list.js
+ */
+
+const { buildDynamoClient, scan } = require("./utils/dynamodb.js");
+const fs = require("fs");
+const path = require("path");
+
+const isLocal = !!process.env.DYNAMODB_URL;
+
+const sectionTableName = isLocal
+  ? "local-section"
+  : process.env.dynamoPrefix + "-section";
+const outputCsvFile = path.resolve(__dirname, "2024-contact-list.csv");
+
+async function handler() {
+  try {
+    console.log("Searching for 2024 modifications");
+
+    buildDynamoClient();
+
+    console.log(`Processing table ${sectionTableName}`);
+    const existingItems = await scan({ TableName: sectionTableName });
+    const filteredItems = filter(existingItems);
+
+    const transformedItems = transform(filteredItems);
+    writeCsv(transformedItems, outputCsvFile);
+
+    console.log(
+      `CSV file generated at ${outputCsvFile} with ${transformedItems.length} records.`
+    );
+
+    return {
+      statusCode: 200,
+      body: "All done!",
+    };
+  } catch (err) {
+    console.error(err);
+    return {
+      statusCode: 500,
+      body: err.message,
+    };
+  }
+}
+
+function filter(items) {
+  return items.filter((item) => item.sectionId === 0 && item.year === 2024);
+}
+
+function transform(items) {
+  return items.map((item) => {
+    const subsections = item.contents.section.subsections?.[0];
+    const parts = subsections.parts?.[0];
+    const questions = parts.questions;
+
+    const stateName = questions[0].answer.entry;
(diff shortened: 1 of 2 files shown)


services/database/scripts/create-contact-list.js
```diff
@@ -0,0 +1,96 @@
...
+const { buildDynamoClient, scan } = require("./utils/dynamodb.js");
+const fs = require("fs");
+const path = require("path");
+
+const isLocal = !!process.env.DYNAMODB_URL;
+
+const sectionTableName = isLocal
+  ? "local-section"
+  : process.env.dynamoPrefix + "-section";
+const outputCsvFile = path.resolve(__dirname, "2024-contact-list.csv");
```
bangbay-bluetiger, line 18: Instead of having it output to the current directory, it would be better if it was a top-level directory. What if we create an `output` folder that we also add to `.gitignore`.

services/database/scripts/create-contact-list.js
```diff
@@ -0,0 +1,96 @@
...
+
+async function handler() {
+  try {
+    console.log("Searching for 2024 modifications");
```
bangbay-bluetiger, line 22: It would be nice if the year was an env variable. Could do something like `const year = process.env.YEAR || 2024` and use the year variable throughout the file.

services/database/scripts/create-contact-list.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+
+function filter(items) {
+  return items.filter((item) => item.sectionId === 0 && item.year === 2024);
+}
+
+function transform(items) {
+  return items.map((item) => {
+    const subsections = item.contents.section.subsections?.[0];
```
bangbay-bluetiger, line 56: Curious if it'll always be the first item in the array.

services/database/scripts/create-contact-list.js
```diff
@@ -0,0 +1,96 @@
...
+    const parts = subsections.parts?.[0];
```
bangbay-bluetiger, line 57: Same. Curious if it'll always be the first item in the array.

services/database/scripts/create-contact-list.js
```diff
@@ -0,0 +1,96 @@
...
+    const questions = parts.questions;
```
bangbay-bluetiger, line 58: Could destructure here: `const { questions } = parts`;

services/database/scripts/create-contact-list.js
```diff
@@ -0,0 +1,96 @@
...
+
+    const stateName = questions[0].answer.entry;
```
bangbay-bluetiger, line 60: Same question about the array order. For these it seems we could filter by `label` to be on the safe side. We could also use destructuring though it may not be as readable:
```
const { answer: { entry: stateName } } = questions.find((q) => q.label === "State or territory name:");
```

bangbay-bluetiger (commented):
Cool work! Some suggestions.

services/database/scripts/create-contact-list.js
JonHolman: Good idea!

services/database/scripts/create-contact-list.js
JonHolman: Good call.

services/database/scripts/create-contact-list.js
JonHolman: I am too.  From what I am seeing, it looks like its hard coded.

services/database/scripts/create-contact-list.js
JonHolman: It looks to me like the data structure is hardcoded.  I'll defer to what the team thinks is best.

services/database/scripts/create-contact-list.js
```diff
@@ -0,0 +1,103 @@
...
+const fs = require("fs");
+const path = require("path");
+
+const isLocal = !!process.env.DYNAMODB_URL;
+
+const sectionTableName = isLocal
+  ? "local-section"
+  : process.env.dynamoPrefix + "-section";
+const outputCsvFile = path.resolve(
+  __dirname,
```
bangbay-bluetiger, line 19: Need a `mkdir` if the folder doesn't exist. Similar to this code:
https://github.com/Enterprise-CMCS/macpro-mdct-carts/blob/44f417f69501e7c204d38c7bfaafba33e873186f/services/uploads/src/antivirus.js#L56-L58

Also, instead of `__dirname`, could alternatively use `process.cwd()` to avoid needing to traverse down:
```
path.join(process.cwd(), "output", `${process.env.year}-contact-list.csv`)
```

services/database/scripts/create-contact-list.js
bangbay-bluetiger: I like to avoid magic numbers if possible. I think at least we should use a `find` for the question level so it's easier to discern if it's really the dataset that we want. Something like:
```
const stateSection = questions.find((q) => q.label === "State or territory name:");
const contactSection = questions.find((q) => q.label === "Who should we contact if we have any questions about your report?");

const stateName = stateSection.answer.entry;
const contactDetails = contactSection.questions;

leave everything else the same...
```

JonHolman: Done, thank you!

services/database/scripts/create-contact-list.js
JonHolman: Updated, thank you @bangbay-bluetiger

bangbay-bluetiger (approved):
:fire:

PR #139933, opened 2025-05-12
CMDCT-4361 - cdk transition sls side DO NOT MERGE

JonHolman:
### Description
This is the changes we need to make to the serverless setup before we convert to cdk.

### Related ticket(s)
CMDCT-4361

---
### How to test
You can double check that we haven't left out any resources that could or should be retained. At a minimum we need Cognito Userpool, most s3 buckets, all Dynamo Tables, and Cloudfront Distribution.
The tests don't pass because this change breaks the app. That's fine when we are in the moment of transition but that's one reason we should never merge this :)

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/output.sh
+++ b/services/output.sh
@@ -1,4 +1,3 @@
-
 set -e
 
 help='This script is run with the format  ./output.sh <target service name> <serverless output variable name> <stage name (optional, default dev)>'
@@ -15,7 +14,7 @@ service=${1}
 output=${2}
 stage=${3}
 
-if [ $output == "url" ]; then
+if [ $output = "url" ]; then
   output="CloudFrontEndpointUrl"
 fi
 
--- a/services/ui/serverless.yml
+++ b/services/ui/serverless.yml
@@ -27,9 +27,7 @@ custom:
   region: ${opt:region, self:provider.region}
   serverlessTerminationProtection:
     stages:
-      - main
-      - val
-      - production
+      - nonsense
   route53HostedZoneId: ${ssm:/configuration/${self:custom.stage}/route53/hostedZoneId, ssm:/configuration/default/route53/hostedZoneId, ""}
   route53DomainName: ${ssm:/configuration/${self:custom.stage}/route53/domainName, ""}
   cloudfrontCertificateArn: ${ssm:/configuration/${self:custom.stage}/cloudfront/certificateArn, ssm:/configuration/default/cloudfront/certificateArn, ""}
@@ -100,16 +98,6 @@ params:
 
 resources:
   Conditions:
-    CreateDnsRecord:
-      Fn::And:
-        - Fn::Not:
-            - Fn::Equals:
-                - ""
-                - ${self:custom.route53HostedZoneId}
-        - Fn::Not:
-            - Fn::Equals:
-                - ""
-                - ${self:custom.route53DomainName}
     CreateCustomCloudFrontDomain:
       Fn::And:
         - Fn::Not:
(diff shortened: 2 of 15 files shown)


services/output.sh
```diff
@@ -15,7 +14,7 @@ service=${1}
 output=${2}
 stage=${3}
 
-if [ $output == "url" ]; then
+if [ $output = "url" ]; then
```
peoplespete, line 17: What's the difference here? Pardon my low shell script knowledge.

services/ui/serverless.yml
```diff
@@ -100,16 +98,6 @@ params:
 
 resources:
   Conditions:
-    CreateDnsRecord:
```
peoplespete, line 103: Love this!

services/ui/serverless.yml
```diff
@@ -314,6 +308,7 @@ resources:
           LogFilePrefix: ${env:LOGGING_BUCKET, ssm:/configuration/${self:custom.stage}/s3/accessLogsPrefix, ssm:/configuration/default/s3/accessLogsPrefix}
     WaflogsUploadBucketPolicy:
```
peoplespete, line 309: So do we import this as well? How come we don't destroy it and have it remade in the cdk so it looks the same. Or was that causing a problem. That feels like maybe the answer I just can't remember.

services/uploads/serverless.yml
```diff
@@ -211,6 +211,7 @@ resources:
                   aws:SecureTransport: false
     AttachmentsBucketPolicy:
       Type: AWS::S3::BucketPolicy
+      DeletionPolicy: Retain
```
peoplespete, line 214: Same question here and below about importing bucket policies.

src/run.ts
```diff
@@ -251,10 +381,73 @@ async function destroy_stage(options: {
...
+    );
+    return;
+  }
+
+  const accountId = await getAccountId();
   await destroyer.destroy(`${process.env.REGION_A}`, options.stage, {
     wait: options.wait,
     filters: filters,
     verify: options.verify,
+    bucketsToSkip: [
```
peoplespete, line 445: This continues to be a very elegant solution to a very big problem 🏅

src/serverless-stage-destroyer.ts
```diff
@@ -0,0 +1,324 @@
...
+export class ServerlessStageDestroyer {
+  public async destroy(
+    region: string,
+    stage: string,
+    options: {
+      filters: Tag[];
+      verify: boolean;
+      wait: boolean;
+      bucketsToSkip?: string[];
+    }
```
peoplespete, line 34:
```suggestion
    options: {
      filters: Tag[];
      verify: boolean;
      wait: boolean;
      bucketsToSkip?: string[];
    } =  = {
    filters: [],
    verify: true,
    wait: true,
    bucketsToSkip: [],
  }
```
I think you can do this to set defaults here. But also 🤷 not sure it's better

peoplespete (dismissed):
A few little comments here and there to think about but STELLAR and could be used immediately if needed.

services/output.sh
JonHolman: `=` is more universal.  `==` only works in bash, while `=` works everywhere.

services/ui/serverless.yml
JonHolman: The bucket or the bucket policy?  We can't delete the bucket because we need to preserve the logs in it.  did you import this on other apps?  If so, can you point me to an example?

services/uploads/serverless.yml
JonHolman: We do import this bucket policy.

PR #139934, opened 2025-05-12
CMDCT-4361 - cdk conversion

JonHolman:
### Description
This swaps out sls infrastructure for cdk

### Related ticket(s)
CMDCT-4361

---
### How to test
Use the app and make sure you can log in

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


.env.tpl
```diff
@@ -11,16 +11,12 @@ stateStatusTableName=local-state-status
 sectionBaseTableName=local-section-base
 sectionTableName=local-section
 stageEnrollmentCountsTableName=local-stg-enrollment-counts
```
peoplespete, line 13: We can remove do this as part of mop up or now since they are no longer used.

deployment/constructs/lambda-dynamo-event.ts
```diff
@@ -58,6 +51,19 @@ export class LambdaDynamoEventSource extends Construct {
               ],
               resources: ["arn:aws:logs:*:*:*"],
             }),
+            new iam.PolicyStatement({
+              effect: iam.Effect.ALLOW,
+              actions: [
+                "dynamodb:DescribeStream",
+                "dynamodb:GetRecords",
+                "dynamodb:GetShardIterator",
+                // "dynamodb:ListShards",
```
peoplespete, line 60: Can we remove this? Or include it? I'd rather to one of those than leave commented out.

deployment/constructs/lambda-dynamo-event.ts
```diff
@@ -74,6 +80,7 @@ export class LambdaDynamoEventSource extends Construct {
       bundling: {
         minify: true,
         sourceMap: true,
+        nodeModules: ["kafkajs"],
```
peoplespete, line 83: Did this cause a new problem that necessitated this addition?

deployment/constructs/lambda-dynamo-event.ts
```diff
@@ -88,6 +95,7 @@ export class LambdaDynamoEventSource extends Construct {
           functionName: this.lambda.functionArn,
           startingPosition: "TRIM_HORIZON",
           maximumRetryAttempts: 2,
+          batchSize: 10,
```
peoplespete, line 98: Looks like elsewhere we've used the default. I'd rather do that even if serverless had a different default.

deployment/constructs/lambda-kafka-event.ts
```diff
@@ -0,0 +1,108 @@
...
+  brokerString?: string;
+  kafkaBootstrapServers: string[];
+  securityGroupId: string;
+  subnets: string[];
+  topics: string[];
+  consumerGroupId: string;
+  stackName: string;
+}
+
+export class LambdaKafkaEventSource extends Construct {
```
peoplespete, line 21: Is this a thing that we're seeing for the first time in an app?
Doesn't feel familiar but does look like something we probably need.

deployment/deployment-config.ts
```diff
@@ -37,7 +41,7 @@ export const determineDeploymentConfig = async (stage: string) => {
     config.secureCloudfrontDomainName = `https://${config.cloudfrontDomainName}/`;
   }
 
-  if (!isLocalStack) {
+  if (!isLocalStack && stage !== "bootstrap") {
```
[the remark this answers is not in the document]
peoplespete: Actually I think you already have them all there.

deployment/stacks/bigmac-streams.ts
```diff
@@ -0,0 +1,134 @@
...
+  props: CreateBigmacStreamsComponentsProps
+) {
+  const {
+    scope,
+    stage,
+    project,
+    vpcName,
+    kafkaAuthorizedSubnetIds,
+    brokerString,
+    kafkaBootstrapServers,
```
peoplespete, line 32: Rather than pass this in let's create it here. Bc of the commonProps use you already have to have brokerString passed in and you don't need kaftkaBootstrapServers elsewhere in parent.

deployment/stacks/bigmac-streams.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  const commonProps = {
+    project,
+    stage,
+    brokerString,
+    stackName: `${service}-${stage}`,
+  };
+
+  // TODO: these permissions are needed for one or both of these functions:
```
peoplespete, line 70: Do we need to do this?

deployment/stacks/bigmac-streams.ts
```diff
@@ -0,0 +1,134 @@
...
+  new LambdaDynamoEventSource(scope, "postKafkaData", {
+    entry: "services/carts-bigmac-streams/handlers/kafka/post/postKafkaData.js",
+    handler: "handler",
+    timeout: Duration.seconds(120),
+    memorySize: 2048,
+    retryAttempts: 2,
+    vpc,
+    vpcSubnets: { subnets: kafkaAuthorizedSubnets },
+    securityGroups: [lambdaSG],
+    environment: {
```
peoplespete, line 127: Since this is the first use of this construct, you could probably move some of the common values into the construct as default.

deployment/constructs/lambda-dynamo-event.ts
JonHolman: yeah. I  didn't consider this to be a fun expedition :)

deployment/constructs/lambda-dynamo-event.ts
JonHolman: so you want to remove this line and just not specify it?

JonHolman: this came up from my painstaking generated cloud formation comparisons

deployment/constructs/lambda-kafka-event.ts
JonHolman: AFAIK this is the first time for this, but yeah. this app definitely needs it.

deployment/constructs/lambda.ts
```diff
@@ -93,7 +85,7 @@ export class Lambda extends Construct {
       bundling: {
         minify: true,
         sourceMap: true,
-        nodeModules: ["jsdom"],
+        // nodeModules: ["jsdom"],
```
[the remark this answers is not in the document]
JonHolman: I would like for more testing to be done before we remove it.

deployment/stacks/deployFrontend.ts
```diff
@@ -1,60 +1,78 @@
 import { Construct } from "constructs";
 import {
-  aws_iam as iam,
   aws_cloudfront as cloudfront,
-  Duration,
+  aws_iam as iam,
   aws_s3 as s3,
   aws_s3_deployment as s3_deployment,
   custom_resources as cr,
+  Duration,
```
peoplespete, line 8: 🥇 For this kind of thing

deployment/stacks/deployFrontend.ts
```diff
@@ -1,60 +1,78 @@
...
+      COGNITO_REDIRECT_SIGNOUT: applicationEndpointUrl + "postLogout",
+      COGNITO_REGION: "us-east-1",
+      COGNITO_USER_POOL_CLIENT_DOMAIN: userPoolClientDomain,
+      COGNITO_USER_POOL_CLIENT_ID: userPoolClientId,
+      COGNITO_USER_POOL_ID: userPoolId,
+      LOCAL_LOGIN: "false",
+      POST_SIGNOUT_REDIRECT: redirectSignout,
+      REACT_APP_LD_SDK_CLIENT: launchDarklyClient,
+      S3_ATTACHMENTS_BUCKET_NAME: s3AttachmentsBucketName,
+      S3_ATTACHMENTS_BUCKET_REGION: "us-east-1",
```
peoplespete, line 70: I think bc we've got this hardcoded (much better) we can rip it out of services/ui/.env.tpl and remove configureLocal.sh entirely.

deployment/stacks/deployFrontend.ts
```diff
@@ -106,14 +124,18 @@ export function deployFrontend(props: DeployFrontendProps) {
       destinationKey: "env-config.js",
       source: path.join("./deployment/stacks/", "env-config.template.js"),
       substitutions: {
-        stage,
+        localLogin: "false",
```
peoplespete, line 127: Rather than substitute let's hardcode it in env-config.template.js

deployment/stacks/imports_included/parent.ts
```diff
@@ -27,5 +34,11 @@ export class ImportsIncludedParentStack extends Stack {
       scope: this,
       stage,
     });
+    createUploadsComponents({
+      scope: this,
+      stage,
+      loggingBucket,
+      isDev: false,
```
peoplespete, line 41:
```suggestion
      isDev,
```
Line 19 hardcodes the value so let's just do it once.

deployment/stacks/imports_included/uploads.ts
```diff
@@ -0,0 +1,65 @@
...
+          s3.HttpMethods.DELETE,
+          s3.HttpMethods.HEAD,
+        ],
+        allowedHeaders: ["*"],
+        exposedHeaders: ["ETag"],
+        maxAge: 3000, // 50 minutes
+      },
+    ],
+    enforceSSL: true,
+    serverAccessLogsBucket: loggingBucket,
```
peoplespete, line 40: I'm assuming it was already wired up like this. If not, let's just put this in the main uploads.ts.

deployment/stacks/parent.ts
```diff
@@ -47,6 +46,8 @@ export class ParentStack extends Stack {
       createApiComponents({
         ...commonProps,
         tables,
+        uploadS3BucketName: "",
+        fiscalYearTemplateS3BucketName: "",
```
peoplespete, line 50: I would have thought we'd want the actual values here.

peoplespete: Ah I see. Let's rip out the use of cognito env variables in authorizer/authenticator like we've done in QMR and SEDS and HCBS so that we don't have to have 2 calls to createApiComponents. It also removes the need for the createAuthRole function.

deployment/stacks/data.ts
```diff
@@ -53,40 +108,36 @@ export function createDataComponents(props: CreateDataComponentsProps) {
...
+      seedTestData: isDev.toString(),
     },
     bundling: {
       commandHooks: {
         beforeBundling(inputDir: string, outputDir: string): string[] {
           return [
-            `mkdir -p ${outputDir}/data/initial_data_load/`,
-            `cp -r ${inputDir}/services/database/data/initial_data_load/* ${outputDir}/data/initial_data_load/`,
+            `mkdir -p ${outputDir}/data/seed/`,
+            `cp -r ${inputDir}/services/database/data/seed/* ${outputDir}/data/seed/`,
```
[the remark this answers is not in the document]
JonHolman: We have it in SEDS as well.  Copying the seed data stuff to the output/build directory.

deployment/stacks/parent.ts
```diff
@@ -95,6 +108,16 @@ export class ParentStack extends Stack {
       userPoolClientId,
       userPoolClientDomain: `${userPoolDomainName}.auth.${Aws.REGION}.amazoncognito.com`,
       customResourceRole,
+      s3AttachmentsBucketName: attachmentsBucket!.bucketName,
+    });
+
+    createBigmacStreamsComponents({
+      ...commonProps,
+      kafkaBootstrapServers: brokerString.split(","),
```
peoplespete, line 116: Same suggestion as somewhere else, to move this kafkaBootstrapServers into the inside of createBigmacStreamsComponents

deployment/stacks/bigmac-streams.ts
JonHolman: I was trying to be consistent with our other constructs and a construct shouldn't be hardcoded to a single use case.  What would you hard code in the construct?

deployment/stacks/ui-auth.ts
```diff
@@ -94,56 +91,65 @@ export function createUiAuthComponents(props: CreateUiAuthComponentsProps) {
...
   const supportedIdentityProviders = [
     cognito.UserPoolClientIdentityProvider.custom(providerName),
+    cognito.UserPoolClientIdentityProvider.COGNITO,
   ];
 
   const appUrl =
     secureCloudfrontDomainName ??
     applicationEndpointUrl ??
-    "https://localhost:3000/";
+    "http://localhost:3000/";
```
peoplespete, line 109: We shouldn't need this line since this file is not used in run local, right?

deployment/stacks/ui-auth.ts
```diff
@@ -94,56 +91,65 @@ export function createUiAuthComponents(props: CreateUiAuthComponentsProps) {
...
       ],
-      callbackUrls: [appUrl],
-      defaultRedirectUri: appUrl,
-      logoutUrls: [appUrl],
+      callbackUrls: [appUrl, "http://localhost:3000/"],
+      logoutUrls: [
+        appUrl,
+        `${appUrl}postLogout`,
+        "http://localhost:3000/",
+        "http://localhost:3000/postLogout",
```
peoplespete, line 132:
```suggestion
```

deployment/stacks/ui-auth.ts
```diff
@@ -94,56 +91,65 @@ export function createUiAuthComponents(props: CreateUiAuthComponentsProps) {
...
       },
       scopes: [
         cognito.OAuthScope.EMAIL,
         cognito.OAuthScope.OPENID,
         cognito.OAuthScope.PROFILE,
       ],
-      callbackUrls: [appUrl],
-      defaultRedirectUri: appUrl,
-      logoutUrls: [appUrl],
+      callbackUrls: [appUrl, "http://localhost:3000/"],
```
peoplespete, line 127:
```suggestion
      callbackUrls: [appUrl],
```

deployment/stacks/bigmac-streams.ts
JonHolman: I'm confused.  It's being passed in via props.  How else is bigmac-streams.ts going to get it?

deployment/stacks/ui.ts
```diff
@@ -50,35 +44,45 @@ export function createUiComponents(props: CreateUiComponentsProps) {
...
+      enableLogging: true,
+      logBucket,
+      logFilePrefix: `AWSLogs/CLOUDFRONT/${stage}/`,
+    };
+  }
 
   const securityHeadersPolicy = new cloudfront.ResponseHeadersPolicy(
     scope,
-    "CloudFormationHeadersPolicy",
+    "CloudFrontHeadersPolicy",
```
peoplespete, line 85: 🍦 for this catch

deployment/stacks/ui.ts
```diff
@@ -50,35 +44,45 @@ export function createUiComponents(props: CreateUiComponentsProps) {
     autoDeleteObjects: true,
     enforceSSL: true,
     blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
-    versioned: true,
-    serverAccessLogsBucket: loggingBucket,
-    serverAccessLogsPrefix: `AWSLogs/${Aws.ACCOUNT_ID}/s3/`,
+    versioned: false,
```
peoplespete, line 47: I put this true for QMR (which doesn't mean it's the right choice :) )
I figured we had to for security hub.

deployment/stacks/uploads.ts
```diff
@@ -0,0 +1,229 @@
...
+        exposedHeaders: ["ETag"],
+        maxAge: 3000, // 50 minutes
+      },
+    ],
+    enforceSSL: true,
+    serverAccessLogsBucket: loggingBucket,
+    serverAccessLogsPrefix: `AWSLogs/${Aws.ACCOUNT_ID}/s3/`,
+  });
+
+  // TODO: one or some of the lambdas need s3:GetObject permissions to this bucket
```
peoplespete, line 56: Is this still a todo or can we rip it out?

deployment/stacks/deployFrontend.ts
```diff
@@ -1,60 +1,78 @@
...
+    userPoolClientId,
+    userPoolId,
   } = props;
 
   const reactAppPath = "./services/ui-src/";
   const buildOutputPath = path.join(reactAppPath, "build");
   const fullPath = path.resolve(reactAppPath);
 
-  execSync("yarn run build", {
+  execSync("CI=false SKIP_PREFLIGHT_CHECK=true yarn run build", {
```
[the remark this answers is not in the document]
JonHolman: We use it to say that this is not running in CI/CD... lol.  But.... but...  it was already there and we're just converting to CDK.  What if removing it breaks something and then we get "CDK broke it"?

deployment/stacks/deployFrontend.ts
```diff
@@ -1,60 +1,78 @@
...
+      API_URL: apiGatewayRestApiUrl,
+      BRANCH_NAME: stage,
+      COGNITO_IDENTITY_POOL_ID: identityPoolId,
+      COGNITO_REDIRECT_SIGNIN: applicationEndpointUrl,
+      COGNITO_REDIRECT_SIGNOUT: applicationEndpointUrl + "postLogout",
+      COGNITO_REGION: "us-east-1",
+      COGNITO_USER_POOL_CLIENT_DOMAIN: userPoolClientDomain,
+      COGNITO_USER_POOL_CLIENT_ID: userPoolClientId,
+      COGNITO_USER_POOL_ID: userPoolId,
+      LOCAL_LOGIN: "false",
```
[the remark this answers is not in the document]
JonHolman: it's the environment variables for when the react app is built.  I'm sure it could use some optimization, but that's not my lane :(

deployment/stacks/uploads.ts
```diff
@@ -0,0 +1,229 @@
...
+  let fiscalYearTemplateBucket: s3.IBucket | undefined;
+  if (!isDev) {
```
peoplespete, line 58: I'm confused about this if. What is the fiscalYearTemplateBucket and why do we only have it in main/val/prod? Like what are the consequences of having it be a part of the ephemeral deployments.

services/app-api/handlers/uploads/createDownloadPsUrl.ts
```diff
@@ -30,7 +30,7 @@ export const getSignedFileUrl = handler(async (event, _context) => {
   const document = results.Items[0];
   // Pre-sign url
   const url = await s3.getSignedDownloadUrl({
-    Bucket: process.env.uploadS3BucketName ?? "local-uploads",
+    Bucket: process.env.uploadS3BucketName,
```
peoplespete, line 33: 👏

deployment/stacks/deployFrontend.ts
JonHolman: how can we hardcode the stage?  the stage could be main, val, prod, jon-cdk, pete-sls, etc.

services/app-api/libs/authorization.ts
```diff
@@ -1,8 +1,6 @@
-import { SSMClient, GetParameterCommand } from "@aws-sdk/client-ssm";
 import { jwtDecode } from "jwt-decode";
 import { IdmRoles, AppRoles, APIGatewayProxyEvent } from "../types";
 import { CognitoJwtVerifier } from "aws-jwt-verify";
```
peoplespete, line 3:
```suggestion
```
Let's just use jwtDecode

deployment/stacks/parent.ts
JonHolman: I wasn't aware of that change.  What is that?  I thought createAuthRole return function was the latest.

deployment/stacks/ui-auth.ts
```diff
@@ -94,56 +91,65 @@ export function createUiAuthComponents(props: CreateUiAuthComponentsProps) {
...
+        "custom:cms_roles": "cmsRoles",
+        "custom:cms_state": "state",
       },
       idpIdentifiers: ["IdpIdentifier"],
     }
   );
 
   const supportedIdentityProviders = [
     cognito.UserPoolClientIdentityProvider.custom(providerName),
+    cognito.UserPoolClientIdentityProvider.COGNITO,
```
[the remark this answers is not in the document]
JonHolman: Tested, you are correct.  Removed.

deployment/stacks/ui-auth.ts
JonHolman: True.  However, the main branch's cognito needs to support localhost:3000.  I can't think of a cleaner way to get there.

deployment/stacks/ui-auth.ts
JonHolman: I just responded, I don't think we should do this.

deployment/stacks/ui-auth.ts
JonHolman: I just responded, I don't think we should do this.

deployment/stacks/ui.ts
```diff
@@ -187,7 +178,7 @@ function setupWaf(
     "CloudfrontWafConstruct",
     {
       name: `${project}-${stage}-ui`,
-      blockByDefault: false,
+      blockByDefault: true,
```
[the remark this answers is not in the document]
JonHolman: I'm not sure what's right.  I was trying to match what was there.  Let's discuss in Orca sync?

deployment/stacks/uploads.ts
JonHolman: needs a little more testing

deployment/stacks/uploads.ts
JonHolman: 🤷‍♂️. matching the as-is

services/app-api/libs/authorization.ts
JonHolman: CognitoJwtVerifier is used.  It's not an unused import.

deployment/stacks/data.ts
```diff
@@ -53,40 +108,36 @@ export function createDataComponents(props: CreateDataComponentsProps) {
               "dynamodb:PutItem",
               "dynamodb:UpdateItem",
               "dynamodb:DeleteItem",
+              // TODO: previous had dynamodb:BatchWriteItem
```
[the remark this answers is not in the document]
ntsummers1: image

deployment/stacks/deployFrontend.ts
ntsummers1: If its not used, feel free to remove it! No worries :)

deployment/constructs/lambda-dynamo-event.ts
peoplespete: I'd match the other cdk implementations at this point. Then if we later decide 10 is better we can specify it as something else for all the apps at once. Not the end of the world sort of thing I just think the more variety we allow now the harder it is to make a standard set of constructs down the road.

deployment/stacks/deployFrontend.ts
peoplespete: I meant hardcoding `localLogin`
Making a substitution to the env-config.template.js only makes sense to me if we don't now the value up front like the apiGatewayRestApiUrl sort of thing. If it is always the same value we can hardcode it in the env-config.template.js like COGNITO_REGION is for example.

deployment/stacks/imports_included/uploads.ts
peoplespete: Yeah my point is that if we are adding serverAccessLogs to this for the first time it's best to import without it and add in the standard deploy stacks after the bucket is safely imported. However if the bucket is already set up with serverAccessLogs it's best to have this here and in the standard deploy stacks as well. That was my thought. But even the thing I'm worried about is not a proven problem so just ignore me on this...

deployment/stacks/parent.ts
peoplespete: We discussed this reconfiguring in a huddle further but let me know if any of it is still confusing.

deployment/stacks/ui-auth.ts
peoplespete: I'm confused because other apps don't have this and they work for local. In this case at least, I'dd guess that the localhost part is never reached for main or ephemeral deploys.

deployment/stacks/ui-auth.ts
peoplespete: I'm just confused then how this is working for HCBS and QMR and SEDS

deployment/stacks/ui-auth.ts
peoplespete: I'm just confused then how this is working for HCBS and QMR and SEDS

deployment/stacks/ui.ts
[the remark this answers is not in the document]
peoplespete: Let's talk about this because HCBS and SEDS have it blockByDefault: false and QMR doesn't have the concept at all. Which is a bit strange since I'd have thought we'd have needed it to be there to get the val and prod (non-vpned) apps accessible.

services/app-api/libs/authorization.ts
peoplespete: Let me know if you're still confused about what I'm suggesting here. I think we talked through it. QMR's master branch has the latest and best setup.

deployment/stacks/ui-auth.ts
JonHolman: they had it at one point.  I wonder if things are going to break.  I think we need to do more testing.

peoplespete (dismissed):
LGTM (assuming you push up the addition of that !isDev if you found moments ago.

peoplespete (dismissed):
NOICE

peoplespete (approved):
LGTM

PR #139939, opened 2025-05-27
Upgrade react to 17

gmrabian:
### Description

- Upgrade React to v17 (which is only a package.json and yarn.lock change thanks to all the prep work done ahead of time!)
- Remove enzyme packages in favor of react-testing-library
  - Enzyme is no longer supported and stopped having official support for react. React-testing-library was already in place, so this mostly swaps over any remaining tests that had not yet been swapped.
  - We had many tests that only rendered a component and expected the test to not fail
    - Where we have more than one test, I removed it
    - Where it's the only test, I modified it

Docs:
[React 17 blog](https://legacy.reactjs.org/blog/2020/10/20/react-v17.html)
[React 17 breaking changes](https://legacy.reactjs.org/blog/2020/08/10/react-v17-rc.html#other-breaking-changes)
[Enzyme to react-testing-library migration guide](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/)

### Related ticket(s)

CMDCT-3818

---
### How to test

- `cd services/ui-src && yarn`
- `cd ../..`
- `./run local`
- App works as expected
- All tests in ui-src pass

---
### Pre-review checklist

### DIFF
--- a/services/ui-src/src/components/layout/FillForm.jsx
+++ b/services/ui-src/src/components/layout/FillForm.jsx
@@ -6,7 +6,6 @@ class FillForm extends Component {
       <div className="fill-form">
         <a
           data-testid="form-action"
-          href="javascript:void(0)"
           onClick={this.props.onClick}
           name={this.props.name}
           title={this.props.title}
--- a/services/ui-src/yarn.lock
+++ b/services/ui-src/yarn.lock
@@ -791,7 +791,7 @@
     json5 "^2.2.3"
     semver "^6.3.1"
 
-"@babel/generator@^7.17.9", "@babel/generator@^7.4.0", "@babel/generator@^7.7.2":
+"@babel/generator@^7.17.9", "@babel/generator@^7.7.2":
   version "7.17.9"
   resolved "https://registry.yarnpkg.com/@babel/generator/-/generator-7.17.9.tgz#f4af9fd38fa8de143c29fce3f71852406fc1e2fc"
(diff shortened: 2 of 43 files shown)


services/ui-src/yarn.lock
```diff
@@ -3700,6 +3700,28 @@
     "@types/babel__core" "^7.20.5"
     react-refresh "^0.14.2"
 
+"@wojtekmaj/enzyme-adapter-react-17@^0.8.0":
```
bangbay-bluetiger, line 3703: When running the app locally, this `yarn.lock` regenerated for me.

bangbay-bluetiger (dismissed):
Looks good to me. Potentially an updated `yarn.lock` didn't get pushed up?

bangbay-bluetiger (approved):

services/ui-src/src/components/layout/FillForm.jsx
```diff
@@ -6,7 +6,6 @@ class FillForm extends Component {
       <div className="fill-form">
         <a
           data-testid="form-action"
-          href="javascript:void(0)"
```
ntsummers1, line 9: This is the one change that has me a little worried. 100% Should be removed. But I know it was put there for a reason. Double checking everything surrounding this is still working just fine.

ntsummers1: Turns out it wasn't put there for a reason? Whole file isn't used for anything but it's own test. Likely legacy code from before we took it over.

ntsummers1 (approved):

PR #139953, opened 2025-06-05
Fix run local for CDK

JonHolman:
### Description
Fix run local for CDK

### Related ticket(s)

CMDCT-

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/deployment/stacks/parent.ts
+++ b/deployment/stacks/parent.ts
@@ -45,7 +45,17 @@ export class ParentStack extends Stack {
       customResourceRole,
     });
 
+    let apiGatewayRestApiUrl = "";
+    let restApiId = "";
+
     if (isLocalStack) {
+      ({ apiGatewayRestApiUrl, restApiId } = createApiComponents({
+        ...commonProps,
+        tables,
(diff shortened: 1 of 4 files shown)


deployment/stacks/parent.ts
```diff
@@ -45,7 +45,17 @@ export class ParentStack extends Stack {
       customResourceRole,
     });
 
+    let apiGatewayRestApiUrl = "";
+    let restApiId = "";
+
     if (isLocalStack) {
+      ({ apiGatewayRestApiUrl, restApiId } = createApiComponents({
```
gmrabian, line 52: nit: if we `return` inside this if statement, we don't need to collect these variables, right? thus we can just initialize them below in the ci/cd call

gmrabian (commented):
suggestion: use something like "placeholder" instead of "dummy"?

peoplespete (approved):
LGTM

gmrabian (approved):

ntsummers1 (approved):
Looks like a AWS error is still popping through, might just have to delete the offending resource.

PR #139977, opened 2025-06-16
Sync package.json to yarn.lock versions

bangbay-bluetiger:
### Description

Through the different ways we make dependency updates (manual, Dependabot, Snyk), sometimes the versions in `package.json` don't get updated although the version is updated in the corresponding `yarn.lock`. This PR updates every `package.json/yarn.lock` pair that has a mismatch. No actual packages have been updated.

---

### How to test

1. Run CARTS locally
2. App should run as normal
3. No `yarn.lock` updates occur from running locally
4. Diff in this PR shows no updates in `version` lines of `yarn.lock` files

### Notes

I wrote a script using `@yarnpkg/lockfile` to find these differences.

### Pre-review checklist

---

#### Review


gmrabian (commented):
To my understanding the purpose of having the `^` symbol in front of the version in a package.json is so that we allow for minor updates in yarn.lock. If we don't want these to ever mismatch, I would suggest removing that symbol (not sure if dependabot would start making those updates itself or not)

bangbay-bluetiger:
@gmrabian From what I've seen in the Dependabot PRs, it updates the package.json to match the yarn.lock for minor version updates. For example, in this [commit](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12250/commits/b58c1d3d260684b512b770650548b32c35465f5f#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519), it updated `axios` to `^1.8.2` in both places.

If a user had updated `^1.8.2` manually, it would have installed the latest minor version `^1.10.0`, like what happened in this [commit](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/pull/12250/commits/f59748116a45728c76f1e00b2b45f5abc9338882). I don't think this is a case of Dependabot mistakingly choosing an older version (`^1.8.2` is three months and four versions ago). It think it is probably treating `^1.8.2` like `1.8.2`.

I don't know if we want to stop using `^`. If we rely on Dependabot for updates, perhaps we'll see mismatches less.

gmrabian (approved):
I'm ok with keeping these in sync, but I also don't want this to become an extra labor for us. Personally I am ok with package and yarn being slightly different (given we are giving it permission to do that with `^`)

PR #139982, opened 2025-06-23
[CMDCT-4759] Fix Unique Key Errors: Repeatable

ntsummers1:
### Description

Fixes the console error coming from Repeatable when creating questions. Turns out some of the questions being passed didn't have id's so this randomly generates one to use!
### Related ticket(s)

CMDCT-4759

---

### How to test

Login as a stateuser
Enter the 2024 report
Navigate to Section 3I
Double check the console to see that there is no longer a React error

### Notes

---

### Pre-review checklist

### DIFF
--- a/services/ui-src/src/components/fields/Repeatable.jsx
+++ b/services/ui-src/src/components/fields/Repeatable.jsx
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { AccordionButton, AccordionPanel } from "@reach/accordion";
+import { v4 as uuidv4 } from "uuid";
 
 import Question from "./Question";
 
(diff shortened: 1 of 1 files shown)


gmrabian (approved):
this makes the warning go away.

broadly speaking, what is the case for questions without ids? is this a mistake we can correct, or a question type we can account for?

ntsummers1:
> this makes the warning go away.
> 
> 
> 
> broadly speaking, what is the case for questions without ids? is this a mistake we can correct, or a question type we can account for?

It's related to how the form generation is setup in carts. There isn't strict typing, and different formats can get passed through. So in this case an actual form question isn't passed, but rather just text could appear. This has caused some problems before, particularly when trying to upgrade anything touching form generation (like cmsds, which when upgraded to v12 spawns fieldset after fieldset). A bit hard to explain in detail in git, but happy to run through it on a screen share :)

PR #139995, opened 2025-07-01
[CMDCT-4757] Fix Unique Key Errors: Fieldset

ntsummers1:
### Description

Looks like the basetemplate doesn't have id's on every thing, so in this particular instance the PRA disclosure had no ID to use as the key. This generates one for now, and in this years upcoming template release would be ideal to update.

### Related ticket(s)

CMDCT-4757

---

### How to test

- Login as stateuser2
- Open up the developer console and clear it
- Open the 2024 CARTS Report
- Ensure you're on the Basic State Information Page:
- Check the console for the error

### Notes

---

### Pre-review checklist

---

### DIFF
--- a/services/ui-src/src/components/fields/Fieldset.jsx
+++ b/services/ui-src/src/components/fields/Fieldset.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
+import { v4 as uuidv4 } from "uuid";
 
 import Question from "./Question";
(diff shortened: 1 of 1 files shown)


gmrabian (approved):
fixes for now but wondering if pra disclosure should be rendered as a field at all 🤔 do we not have a pattern for informational text outside of fields?

britt-mo (approved):

PR #140001, opened 2025-07-02
[CMDCT-4848] Update readme with how ACS and FMAP data is updated

ntsummers1:
### Description

We're looking to document how CARTS pulls in FMAPS from the Federal Register each year so we can use it as a potential guide for how ABCD could do the same. I've gone ahead and update the readme to describe this process, as well as made sure that in the yearly update we have it called out that each year we need to update these numbers. Hopefully this helps clear up how this process works!

### Related ticket(s)

CMDCT-4848

### Pre-review checklist

---


gmrabian (commented):
Great documentation! Easy to follow and broken up into concise sections.

I have preferences for technical writing and I'm including those here as suggestions, along with a few typo fixes. Feel free to take what you like from these and ignore the rest.

gmrabian:
General comment on documentation: when do we want things like this in the repository and when do we want them in confluence? I think having this in the repo makes sense for yearly updates.

cc: @jessabean @bangbay-bluetiger

PR #140045, opened 2025-08-19
[CMDCT-4911] Test Coverage Part 5: Print

ntsummers1:
### Description

Print tests are actually here! Dunno why I've held onto this PR. Its was done a week ago. But here we are! Enjoy some nice test updates and the most mocking I've had to do to make a file work. 

### Related ticket(s)

CMDCT-4911

---

### How to test

cd over into services/ui-src
yarn test --coverage and see that nice green Print coverage :)

### Notes

---

### Pre-review checklist

---

_convert to a different template: [val release](?expand=1&template=val-deployment.md)_ | _[production release](?expand=1&template=production-deployment.md)_

### DIFF
--- a/services/ui-src/src/components/sections/Print.jsx
+++ b/services/ui-src/src/components/sections/Print.jsx
@@ -12,7 +12,6 @@ import Section from "../layout/Section";
 import statesArray from "../utils/statesArray";
 import { loadEnrollmentCounts, loadSections } from "../../actions/initial";
 import { apiLib } from "../../util/apiLib";
-import { v4 as uuidv4 } from "uuid";
 import { Buffer } from "buffer";
 
 const openPdf = (basePdf) => {
@@ -133,12 +132,13 @@ export const Print = () => {
 
   // Check if formData has values
   if (formData !== undefined && formData.length !== 0) {
-    sections.push(<Title urlStateName={stateName} />);
+    sections.push(<Title key={"sectitle-1"} urlStateName={stateName} />);
 
     if (sectionId) {
       // Add section to sections array
       sections.push(
         <Section
+          key={"sec-1"}
           data-testid="print-section"
           sectionId={sectionId}
           subsectionId={subsectionId}
@@ -148,12 +148,10 @@ export const Print = () => {
       );
     } else {
       // Loop through each section to get sectionId
-      /* eslint-disable no-plusplus */
       for (let i = 0; i < formData.length; i++) {
         const sectionId = formData[i].contents.section.id;
 
         // Loop through subsections to get subsectionId
-        /* eslint-disable no-plusplus */
         for (
           let j = 0;
           j < formData[i].contents.section.subsections.length;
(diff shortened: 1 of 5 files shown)


gmrabian (dismissed):

services/ui-src/src/components/sections/Print.jsx
```diff
@@ -13,14 +13,69 @@ import statesArray from "../utils/statesArray";
...
+};
 
 /**
  * Generate data and load entire form based on user information
  *
  * @returns {JSX.Element}
  * @constructor
  */
-const Print = () => {
+export const Print = () => {
```
bangbay-bluetiger, line 78: This is throwing a `Warning: Each child in a list should have a unique "key" prop.`

It comes from needing to add keys to the `sections` rather than the parent div.

Key needed here:

https://github.com/Enterprise-CMCS/macpro-mdct-carts/blob/cd2dd85fd6ac1da1c185e246c260235bac3cd863/services/ui-src/src/components/sections/Print.jsx#L136

and here:

https://github.com/Enterprise-CMCS/macpro-mdct-carts/blob/cd2dd85fd6ac1da1c185e246c260235bac3cd863/services/ui-src/src/components/sections/Print.jsx#L141-L147

and here:

https://github.com/Enterprise-CMCS/macpro-mdct-carts/blob/cd2dd85fd6ac1da1c185e246c260235bac3cd863/services/ui-src/src/components/sections/Print.jsx#L166-L171

`key={uuidv4()}` can be removed:

https://github.com/Enterprise-CMCS/macpro-mdct-carts/blob/cd2dd85fd6ac1da1c185e246c260235bac3cd863/services/ui-src/src/components/sections/Print.jsx#L181

services/ui-src/src/components/sections/Print.test.jsx
```diff
@@ -92,4 +107,85 @@ describe("<Print />", () => {
     const sections = screen.getAllByTestId("print-section");
     expect(sections.length).toBe(1);
   });
+
+  test("does not render sections if formData is empty", () => {
```
bangbay-bluetiger, line 111: I would also add a test that this rendered correctly:

https://github.com/Enterprise-CMCS/macpro-mdct-carts/blob/cd2dd85fd6ac1da1c185e246c260235bac3cd863/services/ui-src/src/components/sections/Print.jsx#L136

In these tests, it renders as `<h1>New York CARTS FY 2025 Report</h1>`. Is that expected?

services/ui-src/src/components/sections/Print.jsx
ntsummers1: Oh good catch! Updated :)

gmrabian (approved):

services/ui-src/src/components/sections/Print.test.jsx
bangbay-bluetiger: The test example uses `state=AL`, so I was wondering if the title was pulling in data correctly. Also, although adding a test for the title doesn't increase coverage, I think we should aim to add more meaningful tests when possible.

PR #140057, opened 2025-08-27
cmdct-5039 - improve heading structure

britt-mo:
from the card: 
If there is no section title, then the wrapping `<h2> `should not be added to the DOM and the `<h3>` further down on the page should be a `<h2>`.

Also removed the disabled autosave button from the top and bottom of the report pages

note:
I didn't find `section title` or `nestedSubsectionTitle` actually being set (besides being null values) so I removed those checks.

### Related ticket(s)

CMDCT-5039
CMDCT-5087
CMDCT-5079
---

### How to test

See that the headers are correctly sequential on every page. Styling should remain the same

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---


ntsummers1 (commented):
Nice work! From what I can tell you're 100% right, the original h2 logic was just never being used. I glanced around and checked out the various report page types just to be sure and only 1 thing stood out but it wasn't even directly related (It was actually a bug that is now properly showing doing to this change 😅). I'm totally cool with removing that and reducing complexity since we are not actually using it.

The one that that is now showing due to this tho is a super minor padding issue: 

Screenshot 2025-08-28 at 12 14 48 PM

This is in Section 5. I guess without that h2 giving its margins the text bumps right up to the top. Ideally we'd pad this out a little and make it show like the original view:

Screenshot 2025-08-28 at 12 15 54 PM

ntsummers1:
This looks great! Wasn't able to find anything amiss and the padding is all sorted. Great job 🍾

ntsummers1 (approved):

gmrabian (approved):

cy-by (approved):
Heading structure appears fixed, on more than just the page suggested! Good job 💪🏻 ✅

PR #140063, opened 2025-08-29
CMDCT-5067 - delete logs after three years

britt-mo:
### Description

All apps should retain cloudfront logs for 30 months or 3 years. Since buckets are versioned we also need to make sure that the lifecycle rule doesn't add a delete marker to the file but actually totally removes it.
I initially set the lifecycle policy to 1 day and commented out the `!isDev` check and confirmed that the logs created after the deployment of my branch were deleted. Then I updated the duration to 1095 days.

### Related ticket(s)

CMDCT-5067

---

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---

### DIFF
--- a/deployment/stacks/ui.ts
+++ b/deployment/stacks/ui.ts
@@ -49,35 +49,48 @@ export function createUiComponents(props: CreateUiComponentsProps) {
   let loggingConfig:
     | { enableLogging: boolean; logBucket: s3.Bucket; logFilePrefix: string }
     | undefined;
-  if (!isDev) {
-    // this bucket is not created for ephemeral environments because the delete of the bucket often fails because it doesn't decouple from the distribution gracefully
-    // should you need to test these parts of the infrastructure out the easiest method is to add your branch's name to the isDev definition in deployment-config.ts
-    const logBucket = new s3.Bucket(scope, "CloudfrontLogBucket", {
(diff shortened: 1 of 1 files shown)


deployment/stacks/ui.ts
```diff
@@ -61,6 +61,26 @@ export function createUiComponents(props: CreateUiComponentsProps) {
...
+      expiration: Duration.days(1095),
+      noncurrentVersionExpiration: Duration.days(1095),
+      abortIncompleteMultipartUploadAfter: Duration.days(1095),
+    });
+
+    logBucket.addLifecycleRule({
+      id: "CleanUpDeleteMarkers",
+      enabled: true,
+      noncurrentVersionExpiration: Duration.days(1095),
     });
```
peoplespete, line 84:
```suggestion
      lifecycleRules: [
        {
          expiration: Duration.days(1095),
          noncurrentVersionExpiration: Duration.days(1095),
        },
      ],
    });
```
I think it's an either/or vs doing it in both ways.

peoplespete (dismissed):
Looks great! I think if you like the suggestion we can add it to all of the other apps. Also fyi I made a slight adjustment to the PR title so it would populate in the Jira ticket (all caps on CMDCT seems to matter for some reason)

peoplespete (approved):
Nice!

gmrabian (approved):

PR #140090, opened 2025-09-16
[CMDCT-5078] Add sortable tables to dashboard pages

bangbay-bluetiger:
### Description

- Create `SortableTable` component with `tanstack`
- Replace table on state and CMS home pages

### Related ticket(s)

CMDCT-5078

---

### How to test

1. Run CARTS locally or use https://d2ookl8cwsz93h.cloudfront.net/
2. Log in as `stateuser2@test.com`
3. Test that table sorts as expected
4. Edit a report and certify and submit it
5. Verify that the status changes and link text changes from `Edit` to `View`
6. Test `Print` link works
7. Log in as `cms.user@test.com`
8. Test that table sorts and filters as expected
9. The report should have a `Uncertify` link
10. Verify uncertifying works
11. Test `View` and `Print` links also work

---

### Pre-review checklist

---

#### Review


ntsummers1 (commented):
Niceeeee. This is awesome stuff! I looked and compared to the mdctdev and couldn't notice a difference, other than this actually handles better on mobile devices which is a win! Theres still some mobile view handling we could do but that can be a separate ticket so I won't call it out here. Sorting worked great and as intended :). 

Nit:
Theres now a slightly bigger gap between the table and All Reports. Comparing against dev where the gap is the same between the filter/clear buttons, All Reports, and the table. Could we tighten that up a little?

ntsummers1 (dismissed):

gmrabian (dismissed):
functionality is good

cy-by (approved):

britt-mo (approved):

PR #140155, opened 2025-09-30
[CMDCT-5208] Bring back local seeding for before 2024 reports and the new 2025 report

ntsummers1:
### Description

We removed the local seeded reports to help ensure no hiccups were happening when generating reports and so that the BO's could review the new 2025 report and compare to the 2024 report in Val. This brings back the local seed cause its honestly a nice QOL for devs.

### Related ticket(s)

CMDCT-5208

---

### How to test

Load up the ephemeral branch and click into any of the reports for stateuser2! You should be able to enter all of them again :).

### Pre-review checklist

---


gmrabian (approved):
🙏

PR #140162, opened 2025-10-01
Fix Header logo focus being off center

ntsummers1:
### Description
This is petty, but I hate how that logo icon always has its focus dangling off and not being aligned with the image. I fixed this. Now I can rest in peace.

Before:
Screenshot 2025-10-01 at 2 36 18 PM

After:
Screenshot 2025-10-01 at 2 35 46 PM

Tiny note, it also bothers me that the focuses on parts of this page are different. I do not need much convincing to fix this, but theres other tickets I need to be doing so this will suffice for now.

### How to test

Load up the app as any user, and then tab your way to the header logo. Should be nice and centered :)

### Notes

---

### Pre-review checklist

### DIFF
--- a/services/ui-src/src/components/layout/Header.jsx
+++ b/services/ui-src/src/components/layout/Header.jsx
@@ -87,7 +87,7 @@ export const Header = () => {
       <header className="header">
         <div className="ds-l-container">
           <div className="ds-l-row header-row">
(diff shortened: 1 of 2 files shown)


gmrabian (approved):

cy-by (approved):
:shipit:

PR #140169, opened 2025-10-06
CMDCT-5191 initial playwright setup and dummy test

gmrabian:
### Description

- Add playwright test folder and initiate with config, dependencies, and dummy test
- Add readme for playwright and commands for running playwright to package.json scripts

There will be more PRs to follow with tests specific to CARTS. For now this is the initial setup. 
  - config almost entirely a copy of the one in MFP
  - README a copy of the one in MFP

I chose to break from the MFP pattern of having cypress and playwright configured in `tests/`. Once we are ready to remove cypress, we can just delete the cypress folder and any scripts that call it. This leaves the playwright folder untouched.

Mostly generated by `yarn create playwright`

[Playwright installation docs](https://playwright.dev/docs/intro)

### Related ticket(s)

CMDCT-5191

---

### How to test

- Pull this branch locally
- `cd tests/playwright && yarn playwright install`
- From root and/or from tests/playwright test these commands:
  - `yarn test:e2e` (runs playwright tests, assumes app is running)
  - `yarn test:e2e-ui` (runs playwright tests, starts app)
  - `yarn test:e2e-stable` (runs playwright tests, assumes app is running (excludes tagged tests (there are none)))
  - `yarn test:e2e-stable-ui` (runs playwright tests, starts app (excludes tagged tests (there are none)))

---

### Pre-review checklist

---


ntsummers1 (approved):

tests/playwright/.gitignore
```diff
@@ -0,0 +1,8 @@
+
```
bangbay-bluetiger, line 1: How do you feel about putting this in the top level at `/tests`? Part of me just thinks it'd be nice to have a file in there.

bangbay-bluetiger (approved):
Non-blocking suggestion.

tests/playwright/.gitignore
gmrabian: 🤔 hmm not opposed but this is playwright specific at the moment and I like that it keeps it isolated from cypress. In MFP we treat `tests/` as root for both but I thought here if we kept them separate until we removed cypress it would be cleaner, then we could always bring everything up a level if we wanted to

PR #140189, opened 2025-10-17
CMDCT-5194 Add state user report playwright test

gmrabian:
### Description

- Add state user playwright test that enters a report
- Refactor playwright tests to set up reusable functions and constants
  - This doesn't have to be the final state of how we organize things, and we can always easily compile these into PageObjects, but I wanted to keep things clean at the start
- Remove redundant home page test
  - It was doing the same thing as the auth setup, so now we just run a11y on homepage
  - Prefer to keep tests minimal and useful as we are building them out

### Related ticket(s)

CMDCT-5194

---

### How to test

- Checkout this branch
- `yarn test:e2e-ui` (runs app and opens playwright ui)
  - If you haven't used playwright in CARTS yet, first: `cd tests/playwright && yarn playwright install`
- Run all tests

---

### Pre-review checklist

---

### DIFF
--- a/tests/playwright/tests/home.spec.ts
+++ b/tests/playwright/tests/home.spec.ts
@@ -1,18 +1,8 @@
-import { test, expect } from "@playwright/test";
-import { e2eA11y } from "../utils/a11y";
+import { test } from "@playwright/test";
+import { e2eA11y } from "../utils";
 
 test.describe("State user home page", () => {
   test.use({ storageState: ".auth/user.json" });
-  test("displays", async ({ page }) => {
-    await page.goto("/");
-    await expect(
-      page.getByRole("heading", { name: "CARTS Developer Login" })
-    ).not.toBeVisible();
-    await expect(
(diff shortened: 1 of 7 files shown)


tests/playwright/tests/home.spec.ts
```diff
@@ -1,18 +1,8 @@
-import { test, expect } from "@playwright/test";
-import { e2eA11y } from "../utils/a11y";
+import { test } from "@playwright/test";
+import { e2eA11y } from "../utils";
 
 test.describe("State user home page", () => {
   test.use({ storageState: ".auth/user.json" });
-  test("displays", async ({ page }) => {
```
gmrabian: tests the same thing as our auth setup. Unless we have a real need to test hitting the home page, we can rely on our setup and other tests.

marcus-myers-co (approved):

ntsummers1 (approved):
ran locally and looking good!

PR #140196, opened 2025-10-27
[CMDCT-5267] Fix tables not calculating imported values

ntsummers1:
### Description

Every odd year (2025, 2023, 2021..) tries to import data from the year prior specifically for Section 3c Parts 5 and 6. The way thats currently handled is at the integer field and text fields. That means nothing else that sits higher in the chain can act on that, such as the synthesize function which is what does all the calculations for the tables that get generated in CARTS. 

So how do we grab data in the synthesize function? 

In CARTS the lastYearsFormData is located in State. So the table can grab that form data from the state and pass it along to the calculations that happen in the synthesize function. Thus we're able to handle any case where the user is using some, all, or none of the imported data.

Heres what that looks like!

https://github.com/user-attachments/assets/e1df66b6-9c4e-4f38-b5c1-3bc798a2b21c

### Related ticket(s)

CMDCT-5267

---

### How to test

Load up the **2024** report and navigate to Section 3c, Part 5. 
For Question 2, click yes or no
Input distinct values you'll remember for a few of the values below
Load up the **2025** report and navigate to Section 3c, Part 5. 
For Question 2, answer the same way as you did above
Input distinct a distinct value you'll remember.
Scroll to the bottom of the 2025 tables and note theres values that were inputted, as well as any values to may have overwritten in the 2025 report!

---

### Pre-review checklist

---

#### Review

### DIFF
--- a/services/ui-src/src/util/synthesize.test.js
+++ b/services/ui-src/src/util/synthesize.test.js
@@ -41,39 +41,82 @@ const state = {
       id: "section1",
       items: [
         {
-          id: "item0",
+          id: "2025-01-a-01-01-a",
           answer: { entry: "0" },
         },
         {
-          id: "item1",
+          id: "2025-01-a-01-01-b",
           answer: { entry: "1" },
         },
         {
-          id: "item2",
+          id: "2025-01-a-01-01-c",
           answer: { entry: "2" },
         },
         {
-          id: "item3",
+          id: "2025-01-a-01-01-d",
           answer: { entry: "3" },
         },
         {
-          id: "item4",
+          id: "2025-01-a-01-01-e",
           answer: { entry: "4" },
         },
         {
-          id: "item5",
+          id: "2025-01-a-01-01-f",
           answer: { entry: "5" },
         },
         {
-          id: "item6",
+          id: "2025-01-a-01-01-g",
           answer: { entry: null },
         },
         {
-          id: "item7",
+          id: "2025-01-a-01-01-h",
           answer: { entry: "abc" },
         },
         {
-          id: "item8",
+          id: "2025-01-a-01-01-i",
+          answer: { entry: "1,001" },
+        },
+      ],
+    },
+  ],
+  lastYearFormData: [
+    {
+      id: "section1",
+      items: [
+        {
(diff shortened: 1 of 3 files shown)


services/ui-src/src/util/synthesize.test.js
```diff
@@ -490,146 +534,148 @@ describe("synthesize()", () => {
...
-          state.allStatesData,
-          state.global.stateName,
-          state.stateUser.abbr,
-          state.enrollmentCounts.chipEnrollments,
-          state.formData,
-          true // printView
-        );
-        expect(out).toEqual({ contents: "<11" });
-      });
+    test("returns empty string when numerator is blank", () => {
```
ntsummers1: Nothing changed here other than the order of tests changed. SumAndPercentage was a accidentally a child test group inside numberAndPercentage so I broke it out separately.

services/ui-src/src/util/synthesize.test.js
```diff
@@ -490,146 +534,148 @@ describe("synthesize()", () => {
...
-          },
-          state.allStatesData,
-          state.global.stateName,
-          state.stateUser.abbr,
-          state.enrollmentCounts.chipEnrollments,
-          state.formData
-        );
-        expect(out).toEqual({ contents: "" });
-      });
+    test("numberAndPercentage uses lastYearFormData for odd year and missing values", () => {
```
ntsummers1: This is the 1 of 2 new tests thats actually useful to review

services/ui-src/src/util/synthesize.test.js
```diff
@@ -639,6 +685,51 @@ describe("synthesize()", () => {
...
+        state.allStatesData,
+        state.global.stateName,
+        state.stateUser.abbr,
+        state.enrollmentCounts.chipEnrollments,
+        state.formData
+      );
+      expect(out).toEqual({ contents: "" });
+    });
+
+    test("sumAndPercentage uses lastYearFormData for odd year and missing values", () => {
```
ntsummers1: This is the 2nd of 2 useful tests that got added to this file

gmrabian (approved):

PR #140200, opened 2025-10-30
CMDCT-4939 Add skip nav

cy-by:
### Description

Add skip nav link to the top of every page.

### Related ticket(s)
CMDCT-4939

---

### How to test

1. Run local or open [deployed env](https://d1ywwym0neyxig.cloudfront.net)
2. Once the home page is loaded, press the `Tab` key
3. Focus should move to the new skip nav link, which will appear on focus
4. Press `enter` to move focus to `<main>`
5. Press `tab` again to confirm that the next interactive element that gets focus is the next interactive element in the `<main>` section.

This can be repeated on any other page.

#### Video example on home page
https://github.com/user-attachments/assets/f535b10b-cac1-4f47-9402-938789856548

### DIFF
--- a/services/ui-src/src/components/layout/SkipNav.jsx
+++ b/services/ui-src/src/components/layout/SkipNav.jsx
@@ -1,0 +1,18 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export const SkipNav = ({ id, href = "#main-content", text }) => {
+  return (
+    <a id={id} href={href} className="skip-nav">
+      {text}
+    </a>
+  );
+};
+
+SkipNav.propTypes = {
+  id: PropTypes.string.isRequired,
+  href: PropTypes.string,
+  text: PropTypes.string.isRequired,
+};
+
+export default SkipNav;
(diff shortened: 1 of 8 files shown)


services/ui-src/src/styles/_skipNav.scss
```diff
@@ -0,0 +1,22 @@
+@use "variables";
+
+/**********************
+  SKIPNAV
+**********************/
+.skip-nav {
```
cy-by: Most of these values were taken from the MCR/CMS DS skip nav.

I aligned to the SCSS variables in this project as much as I could.

services/ui-src/src/components/layout/SkipNav.jsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export const SkipNav = ({ id, href = "#main-content", text, ...props }) => {
```
gmrabian: nit: if we don't have a use case for `...props` right now I would remove it

super nit: we could [define prop types](https://github.com/Enterprise-CMCS/macpro-mdct-carts/blob/main/services/ui-src/src/components/layout/UploadComponent.jsx#L292-L294)

gmrabian (dismissed):
Looks good! Tested on a few pages, including Print.

gmrabian (approved):

PR #140212, opened 2025-11-13
Nit: Remove semicolon from jsx on local login

cy-by:
### Description

This was driving me crazy 🥴

There was a `;` in the `.jsx` in `App.jsx` that was loaded in the DOM on the local login page.

CARTS Main has semicolon as child

### How to test

1. Start local dev or navigate to the deployed env !!!
2. Check that there is no visual `;` on the page


bangbay-bluetiger (approved):
Nice find!

PR #140225, opened 2025-12-01
only install chromium for playwright ci tests

gmrabian:
### Description

Only install chromium browser since that's all we test with

Doesn't save that much time but 🤷 

---
### How to test

- Playwright tests pass
- Only chromium installs (check logs)


ntsummers1 (approved):
Any time saved is a good time 👍

peoplespete (approved):
🦖

PR #140253, opened 2026-01-06
Enforce a couple package versions for sec vulns

ntsummers1:
### Description

Got some security vulnerabilities tagged by Git. Pretty easy dev-dep updates on these so nothing touching user facing code. Should pop these in :).

Part 2 update: Cypress version got literally updated 1 hour after I updated the version. Soooo we're now on their most recent one as it was updated to fix the sec vulnerability now. 

Other edits were to remove old resolutions that are no longer required, and to enforce the latest security updates on a couple packages while we wait for their parent to be updated.

---
### How to test

---
### Pre-review checklist

### DIFF
--- a/services/ui-src/package.json
+++ b/services/ui-src/package.json
@@ -108,7 +108,7 @@
     }
   },
   "resolutions": {
-    "react-error-overlay": "6.1.0",
-    "tar-fs": "3.1.1"
+    "js-yaml": "3.14.2",
+    "glob": "10.5.0"
   }
 }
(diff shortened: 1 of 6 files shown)


services/ui-src/package.json
```diff
@@ -108,7 +108,7 @@
     }
   },
   "resolutions": {
-    "react-error-overlay": "6.1.0",
-    "tar-fs": "3.1.1"
+    "js-yaml": "3.14.2",
+    "glob": "10.5.0"
```
gmrabian: the [glob alert](https://github.com/Enterprise-CMCS/macpro-mdct-carts/security/dependabot/967) was in app-api/ so may need to move this there (unless they both need it?)

ntsummers1:
Trying to get this to work has uncovered that this is more than just a 1 or 2 dependency fix. Attempting to resolve these will result in multiple other dependencies needed to be forced up and cause outlying issues. Going to push this back for a little bit to see if the main packages causing these sec problems get updated soon. Will keep an eye on this.

PR #140346, opened 2026-04-28
Fix printFormURL not validating print url to prevent xss

ntsummers1:
### Description

We received this warning from SNYK about our printFormUrl function:
image

It reads: "Unsanitized input from the document location flows into a React dynamic 'href' attribute, where it is used to dynamically construct the HTML page on client side. This may result in a DOM Based Cross-Site Scripting attack (DOMXSS)

I've gone ahead and updated the whole FormActions function to use regex to validate out each individual parameter that could have been passed (State Id, Section ID, etc.) and then build out a suitable link for us to use. If anyone updates the URL to provide a different ID for those values, it'll clean it and send the user to the /print page to prevent such an attack.

Example Video of the functionality still working after these changes were made:

https://github.com/user-attachments/assets/ece5cf9a-aaa5-492a-80e0-2893933c07ca

### Related ticket(s)

CMDCT-5874

---
### How to test

- Load up a report as a state user
- Scroll to the bottom of the page and click the print button
- Click print section or full form. Both should work!
- Verify the print view comes up
- Go back to the form and pull up a section like 3c (We build links differently if its in a subsection)
- Scroll to the bottom of the page and click the print button
- Click print section or full form. Both should work!
- Verify the print view comes up
- Log out as a state user and log back in as an admin
- Repeat the steps above and you should see everything still working!

---
### Pre-review checklist

---
#### Review


services/ui-src/src/components/layout/FormActions.jsx
```diff
@@ -18,38 +27,36 @@ const FormActions = () => {
...
+    safeFormYear && routeSectionOrdinal
+      ? `${safeFormYear}-${routeSectionOrdinal}`
+      : null;
+
+  const subsectionMarker =
+    routeSectionOrdinal === SECTION_THREE_ORDINAL
+      ? (routeSubsectionMarker ?? DEFAULT_SUBSECTION_MARKER)
+      : DEFAULT_SUBSECTION_MARKER;
+
+  const subsectionId = sectionId ? `${sectionId}-${subsectionMarker}` : null;
```
ntsummers1, line 59: Hopefully this reads a lot easier. I've also abstracted a lot of the work here over to the new printUrlHelpers file

services/ui-src/src/components/layout/FormActions.jsx
```diff
@@ -89,37 +96,9 @@ const FormActions = () => {
 
   /**
    * Generates the URL to print a form.
-   * @param {object} currentUser - the current user object
    * @param {string} formYear - the year associated with the report
    * @return {string} The URL string
    */
-  const printFormUrl = (
```
ntsummers1, line 96: See the printURLHelpers file for where this went!

services/ui-src/src/components/layout/FormActions.test.jsx
```diff
@@ -119,10 +154,10 @@ describe("<FormActions />", () => {
 
     render(formActions);
     const printShowButton = screen.getByTestId("print-show");
-    fireEvent.click(printShowButton);
 
     // Check clear
     act(() => {
+      userEvent.click(printShowButton);
```
JonHolman: this needs an await?

ntsummers1: So I admit I wasn't sure on this. Wrapping it in the act call makes everything inside synchronous from my understanding. Like if I put just an await on the userEvent it'll yell at me that its only allowed in async functions. I could make the act async to fix that. But I wasn't sure about what that might truly change compared to this setup 🤔

bangbay-bluetiger: `userEvent` calls are always async so this will need an await. I'm not certain that it needs to live inside the `act` though.

ntsummers1: Updated with an await, as well as moved it out of the act 😄

ntsummers1: Updated with an await, as well as moved it out of the act 😄

bangbay-bluetiger (dismissed):

britt-mo (dismissed):
Printing worked as expected

britt-mo (approved):

PR #140354, opened 2026-05-11
[CMDCT-5268] Stop constantly calling AWS when clicking fields

ntsummers1:
### Description

In the past we've been having intermittent tier 3 alarms over the consumed write capacity units for DynamoDB. Based on our discussions, one of the ways to help reduce the frequency of this alarm is to reduce the amount of writes all fields are doing. On focus, entering data, and leaving focus triggers a 1.2 mb write of the whole report. This happens regardless of any actual data input. These changes can get queued in a debounce function, but that function has a timer of 300ms, hence the consistent 300ms write requests we’re seeing that keep getting fired as a user clicks/tabs through the page. 

This PR fixes that insane amount of writes! It turns out we had an overzealous onHandleClick function that would run any time a user clicked in, changed a value, or lost focus. This was meant for checkboxes and radio fields to help capture their checked and unchecked states but bled over into every other field. Thus, now checkboxes and radio fields correctly capture their onHandleClick, while other fields no longer submit endless streams of AWS writes while perusing the report! This should also save some money with AWS :) 

I've also cleaned up a few other things:
- Fixed double-dispatch on checkbox clicks. Previously, clicking a checkbox triggered 2 calls with the first being an empty string. This was causing two Redux dispatches per click and the second call was masking the first. Now only onChange fires once with the actual value of the checkbox.
- Radio and Checkbox now own their onChange/dispatch logic.
- I also fixed checkbox's null handling. If you checked a checkbox you'd end up with the array `[null, 'checkedvalue']`. Null should never be in that array in the first place. Null is there only if a user has every checkbox unchecked. Now checkboxes will have a clean standard output of `null` (If never selected anything), `['checkedvalue1']` (If you've selected a value), or `[]` (If you previously selected an option, but unchecked everything after). We support null here for legacy data purposes.
- Radio buttons now handle like radio buttons. If you select a radio option, that group will now properly require one option to be selected at all times.

Demo of this working in action:

https://github.com/user-attachments/assets/7751d8fc-501e-415d-9eca-800ab3a2373a

### Related ticket(s)

CMDCT-5268

---
### How to test

- Load up any report in the deployed instance
- Play around with all the different fields. The change here related to the Question field which is used by just about ever form field rendered on the page. Users should be able to enter and save data, click in and out of fields without triggering a write, and things like fileUpload should still work
- One thing to note is the data being sent. If you checkout the report payload, you'll no longer see `[null, 'combo']` as a potential value. You'll only see `['combo']` like it should have been in the first place :). 
- Now see how it used to be!
- Run main locally
- Click in and out of a field and watch the calls fly
- Play with the checkboxes and see the above data issue in full view

---
### Pre-review checklist

### DIFF
--- a/services/ui-src/src/components/fields/Checkbox.jsx
+++ b/services/ui-src/src/components/fields/Checkbox.jsx
@@ -9,7 +9,7 @@ const Checkbox = ({ onChange, question, ...props }) => {
       ? selected.filter((v) => v !== value)
       : [...selected, value];
 
-    onChange({ target: { name, value: updated.length > 0 ? updated : null } });
+    onChange({ target: { name, value: updated.length > 0 ? updated : [] } });
   };
 
   const radioButttonList = question.answer.options.map(
--- a/services/ui-src/src/components/fields/Checkbox.jsx
+++ b/services/ui-src/src/components/fields/Checkbox.jsx
@@ -2,18 +2,14 @@ import React from "react";
 import PropTypes from "prop-types";
 
 const Checkbox = ({ onChange, question, ...props }) => {
-  const value = Array.isArray(question.answer.entry)
-    ? question.answer.entry
-    : [question.answer.entry];
+  const selected = [question.answer.entry].flat().filter((v) => v != null);
 
-  const change = ({ target: { name, value: newValue } }) => {
-    const index = value.indexOf(newValue);
-    if (index !== -1) {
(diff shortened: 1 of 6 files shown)


bangbay-bluetiger (dismissed):

britt-mo (approved):
Looked good in the deployed env!

gmrabian (approved):

PR #140359, opened 2026-05-22
Fix whitescreen bug when connecting to application

ntsummers1:
### Description

You've seen it. The white screen when you load up one of our apps requiring you to hard refresh the page to get your to the sign-in page. Well no longer! This multi-year bugs reign is at an end.

The issue, at least from how I understand it, is that users returning to our apps after ~24h saw a blank white screen because their refresh token had expired but stale auth state was still cached in localStorage. Our authentication process now calls `signOut({ global: false })` before redirecting us and signing in. This wipes the expired tokens and stale values from localStorage, so the authentication flow starts from the beginning again. Should mention, that signOut call is only made when a user’s auth is already failing so this will not impact users who have valid/fresh credentials.

### Related ticket(s)

CMDCT-5842

---
### How to test

- Try signing into there, then close the browser tab and check back like 1 day + 30 mins from now (Can be any time after 24 hours). You also need to make sure you don't open anything else related to CARTS during this timeframe as that'll reset your credentials so a weekend is a good time to test! 

---
### Pre-review checklist

### DIFF
--- a/services/ui-src/src/App.jsx
+++ b/services/ui-src/src/App.jsx
@@ -12,10 +12,11 @@ import { fireTealiumPageView } from "./util/tealium";
 import "font-awesome/css/font-awesome.min.css";
 import "./styles/app.scss";
 import { LocalLogins } from "./components/sections/login/LocalLogins";
+import { Button } from "@cmsgov/design-system";
 
 function App() {
   const { pathname, key } = useLocation();
-  const { user, showLocalLogins, loginWithIDM } = useUser();
+  const { user, showLocalLogins, loginWithIDM, authError } = useUser();
 
   // fire tealium page view on route change
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


stacy-coforma (dismissed):
tested this with Nick and works like a charm - pretty sure I am not a codeowner on this repo but approving anyway 😅

BearHanded (dismissed):
Tested after 24.5 hrs and it appears to sidestep the white screen. Code looks sane.

We might want to double check the IDM portion works after a dev merge, but functionally it looks like it just tucks it into a nice catch and error display.

ntsummers1:
Should merge this after the val deploy goes off. Not needed in the upcoming prod release

karla-vm (approved):

bangbay-bluetiger (approved):

PR #140360, opened 2026-05-22
cmdct-6064 - fix currency styling issue on print

britt-mo:
### Description
from the card:

> Currency fields are currently working as expected on form pages. They take user input and display dollar signs on the left of the input. However, if a user goes to the print page, and then clicks print form again, Doc Raptor will generate a PDF without the first number in the field. 

When users clicked the Print button, `getPdfFriendlyDocument()` caused the input field numbers to shift in the UI and would shift down and to the left in the PDF. This PR saves the original input styles before modification, captures the HTML snapshot, then restores styles immediately. This ensures the PDF gets the modified HTML while the page UI returns to normal.

Has been reviewed by product and design

### Related ticket(s)

CMDCT-6064

---
### How to test
Enter a report, go to `section 5` and modify some of the currency fields with different values. 
Go to the bottom of the page and click `print`, I'd just do that section so you don't have to scroll a lot.
See that the currency fields look the same here. Click the `print` button at the top of the page, see that the currency field content doesn't shift after clicking.
Confirm that the currency fields in the PDF are displaying the numbers with correct styling-- numbers shouldn't be shifting to the left or getting covered by the `$`

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/sections/Print.jsx
+++ b/services/ui-src/src/components/sections/Print.jsx
@@ -30,9 +30,20 @@ export const getPdfFriendlyDocument = async () => {
   const html = document.querySelector("html");
   html.querySelector("noscript")?.remove();
 
+  // Save original styles before modifying
+  const inputStyles = [];
+
   document.querySelectorAll("input").forEach((element) => {
     if (element.type === "text") {
+      inputStyles.push({
+        element,
+        height: element.style.height,
+        width: element.style.width,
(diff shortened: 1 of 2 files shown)


karla-vm (approved):
NICE! ✨

ntsummers1 (approved):
Looking fantastic!

PR #140361, opened 2026-05-27
clone the document for print, prevent mutation of live DOM

britt-mo:
### Description

`getPdfFriendlyDocument()` in `Print.jsx` was directly mutating the live DOM. It called `element.remove()` on every `<button>` without `title="Print"`. Since it happened outside React's virtual DOM, the button was gone permanently until a page refresh. It now clones the entire document with `cloneNode(true)` [doc here](https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode) first and performs all mutations (button removal, input styling, etc) on the clone. The content sent to the PDF API is derived from the clone, leaving the live DOM untouched. 
This change made my previous addition of `inputStyles` save/restore logic unnecessary.

### Related ticket(s)

CMDCT-6065

---
### How to test
Login (using stateuser2) and enter a report. 
Scroll to the bottom of the page and select `print`, choose either the section or the entire report. 
On the next page click the `print` button at the top of the page and see that `My Account` in the header doesn't disappear anymore, and still is not present on the next page that opens to print.

You can also verify that this maintains my changes for the styling of input fields if you fill out some of the currency fields in section 5.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/sections/Print.jsx
+++ b/services/ui-src/src/components/sections/Print.jsx
@@ -50,8 +50,8 @@ export const getPdfFriendlyDocument = async () => {
     clonedHtml.querySelector("head").prepend(base);
   }
 
-  const htmlString = clonedHtml
-    .outerHTML.replaceAll(`’`, `'`)
+  const htmlString = clonedHtml.outerHTML
+    .replaceAll(`’`, `'`)
     .replaceAll(`‘`, `'`)
     .replaceAll(`”`, `"`)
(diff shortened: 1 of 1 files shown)


ajaitasaini (approved):
tested in cloudfront link and it's working!! Still seeing the My Account button on the page + the PDF is generating in a different tab

karla-vm (approved):

